PR #34, opened 2019-01-23
Add 32-bit QEMU Virt support

alistair23:
The QEMU virt machine supports both 32-bit and 64-bit RISC-V machines. This patch series adds support for the 32-bit machines.

At the moment with this patch set I can build and boot openSBI. The kernel then fails to boot (only runs a handful of instructions) but I'm not sure why.


alistair23:
V3 pushed.  Now OpenSBI allows us to boot 32-bit Linux.

Linux 5.0  crashes in userspace, but the same crash occur with BBL, so it's a kernel bug not an openSBI bug.

alistair23:
Reopening this as a separate PR from my public fork https://github.com/riscv/opensbi/pull/50

PR #45, opened 2019-02-04
riscv_asm.h: Use CSR_<FOO> instead of <foo> for csr_read()

olofj:
Some toolchains might not have all the CSRs available (as seen with
GCC 7.2). So, instead use the defined CSR_ values.

I think we might want to move over all uses to these instead over time?


avpatel (approved):
Looks good to me.

PR #46, opened 2019-02-04
Makefile: Don't rely on "echo -n"

olofj:
Turns out it doesn't behave as expected on MacOS, it doesn't honor
-n and echoes that to the file instead. Add a slash and just let the
newline be there instead.


avpatel (approved):
LGTM. Thanks for trying out compilation on MacOS

PR #48, opened 2019-02-05
Fix build errors

andreas-schwab:


Makefile
```diff
@@ -8,9 +8,9 @@
 #
 
 # Select Make Options:
-# o  Do not use make's built-in rules and variables
+# o  Do not use make's built-in rules
 # o  Do not print "Entering directory ...";
-MAKEFLAGS += -rR --no-print-directory
+MAKEFLAGS += -r --no-print-directory
```
avpatel: Let's keep the "-R" because we have "CC ?= " below.

If "-R" is removed and there is not native compiler then CC will be set to "" instead of being unset.

andreas-schwab:
On Feb 05 2019, Anup Patel <notifications@github.com> wrote:

> Let's keep the "-R" because we have "CC ?= " below.

That's exactly what -R breaks.

avpatel:
No issues in cross-compilation so merging this PR.

PR #50, opened 2019-02-07
Add 32-bit QEMU Virt support

alistair23:
The QEMU virt machine supports both 32-bit and 64-bit RISC-V machines. This patch series adds support for the 32-bit machines.

### DIFF
--- a/platform/qemu/virt/config.mk
+++ b/platform/qemu/virt/config.mk
@@ -30,10 +30,12 @@ PLATFORM_SYS_CLINT=y
 # Blobs to build
 FW_TEXT_START=0x80000000
 FW_JUMP=y
-FW_JUMP_ADDR=0x80200000
+# This needs to be 4MB alligned for 32-bit support
+FW_JUMP_ADDR=0x80400000
 FW_JUMP_FDT_ADDR=0x82200000
 FW_PAYLOAD=y
-FW_PAYLOAD_OFFSET=0x200000
+# This needs to be 4MB alligned for 32-bit support
+FW_PAYLOAD_OFFSET=0x400000
 FW_PAYLOAD_FDT_ADDR=0x82200000
 
 # External Libraries to include
--- a/platform/qemu/virt/config.mk
+++ b/platform/qemu/virt/config.mk
@@ -9,8 +9,17 @@
 
 # Compiler flags
 platform-cppflags-y =
-platform-cflags-y =-mabi=lp64 -march=rv64imafdc -mcmodel=medany
-platform-asflags-y =-mabi=lp64 -march=rv64imafdc -mcmodel=medany
+# If we know the compillers xlen use it below
+ifeq ($(OPENSBI_CC_XLEN), 32)
+	platform-cflags-y =-mabi=ilp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
+	platform-asflags-y =-mabi=ilp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
+endif
+ifeq ($(OPENSBI_CC_XLEN), 64)
+	platform-cflags-y =-mabi=lp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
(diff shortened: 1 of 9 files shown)


atishp04:
Let's not merge it until we fix the kernel issue.

platform/qemu/virt/config.mk
```diff
@@ -9,8 +9,8 @@
 
 # Compiler flags
 platform-cppflags-y =
-platform-cflags-y =-mabi=lp64 -march=rv64imafdc -mcmodel=medany
-platform-asflags-y =-mabi=lp64 -march=rv64imafdc -mcmodel=medany
+platform-cflags-y =-mcmodel=medany
+platform-asflags-y =-mcmodel=medany
```
avpatel: The OpenSBI is RISC-V specific and generic library under lib/ directory expects 'I', 'M', 'A', 'F', and 'D' extension. The OpenSBI dependency on 'C' extension is optional.

I think having explicit -mabi, -march, and -mcmodel is good because it ensures toolchain being used has required features and code does not use any other RISC-V extension.

Instead this change, let's update config.mk of all platforms and move -mabi, -march, and -mcmodel in top-level Makefile. The top-level Makefile will give two options PLATFORM_RISCV_XLEN and PLATFORM_RISCV_C_EXT.

If PLATFORM_RISCV_XLEN=32 then -mabi=lp32 -march=rv32imafd -mcmodel=medany
If PLATFORM_RISCV_XLEN=64 then -mabi=lp64 -march=rv64imafd -mcmodel=medany

If PLATFORM_HAS_C_EXT=y then "c" is appended to -march

By default, PLATFORM_RISCV_XLEN=64 and PLATFORM_RISCV_C_EXT=y (if not set).

Both PLATFORM_RISCV_XLEN and PLATFORM_RISCV_C_EXT can be set in config.mk or passed as command-line parameter to top-level makefile.

Suggestions???

alistair23:
>           Let's not merge it until we fix the kernel issue.

I'm trying to bisect now, but it's unrelated to this so I think it's still worth merging this in the meantime.

platform/qemu/virt/config.mk
alistair23: This is not dynamic so the virt machine won't build for 32-bit RISC-V without code modifications. Whatever we do needs to be dynamic and adjust based on the compiler.

avpatel: Okay, how about this.

When PLATFORM_RISCV_XLEN and PLATFORM_RISCV_C_EXT are not defined by platform config.mk, we set these variable in Makefile by looking at toolchain capabilities.

This way platforms have a way to enforce particular PLATFORM_RISCV_XLEN and PLATFORM_RISCV_C_EXT. If platforms such as QEMU virt/sifive_u don't want to enforce then top-level Makefile do best guess based on toolchain capabilities.

platform/qemu/virt/config.mk
```diff
@@ -21,10 +21,10 @@ PLATFORM_SYS_CLINT=y
 # Blobs to build
 FW_TEXT_START=0x80000000
 FW_JUMP=y
-FW_JUMP_ADDR=0x80200000
+FW_JUMP_ADDR=0x80400000
```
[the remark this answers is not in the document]
alistair23: Ok, I updated the template as well but I'll update this

alistair23:
> > > ```
> > >       Let's not merge it until we fix the kernel issue.
> > > ```
> > 
> > 
> > I'm trying to bisect now, but it's unrelated to this so I think it's still worth merging this in the meantime.
> 
> Once we merge, somebody will start testing it and create issues that we already know. We have this PR. If 32 bit support is urgent is anybody else, they can always cherry-pick this PR for testing. Thoughts ?

True, but there are bugs in openSBI that should be fixed (even if the kernel doesn't work). There is also no 32-bit glibc support so I don't think people expect too much in 32-bit RISC-V.

platform/qemu/virt/config.mk
alistair23: That is more or less what we do now. If the platform hard codes the abi and arch then we will use that, if not it is up to the compiler. I can make it more modular but it will still be the same effect we have now.

alistair23:
V4 pushed. Both 32 and 64 bit Linux boot in the QEMU virt machine. We now support dynamically determining the xlen as well.

PR #53, opened 2019-02-08
Makefile updates

avpatel:


platform/qemu/sifive_u/config.mk
```diff
@@ -9,8 +9,17 @@
...
+ifeq ($(OPENSBI_CC_XLEN), 32)
+	platform-cflags-y =-mabi=ilp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
+	platform-asflags-y =-mabi=ilp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
+endif
+ifeq ($(OPENSBI_CC_XLEN), 64)
+	platform-cflags-y =-mabi=lp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
+	platform-asflags-y =-mabi=lp$(OPENSBI_CC_XLEN) -march=rv$(OPENSBI_CC_XLEN)imafdc
+endif
+platform-cflags-y +=  -mcmodel=medany
+platform-asflags-y += -mcmodel=medany
```
damien-lemoal: All these changes are being removed by the next patch. Is this really necessary ? Not a big deal though.

avpatel: Yes, these changes do overlap with next patch. I will drop these changes.

avpatel:
Done. Updated PATCH2.

avpatel:
Done. Updated PATCH2.

Makefile
```diff
@@ -119,10 +119,34 @@ DTC		=	dtc
...
 OPENSBI_CC_XLEN = `expr substr \`$(CC) -dumpmachine\`  6 2`
 
+# Setup platform XLEN, ABI, ISA and Code Model
+ifndef PLATFORM_RISCV_XLEN
+ifeq ($(OPENSBI_CC_XLEN), 32)
+PLATFORM_RISCV_XLEN = 32
+else
+PLATFORM_RISCV_XLEN = 64
+endif
+endif
```
olofj: This is hard to read. I think there are better ways to write it, but I don't know make quite well enough for it. But I do know it's possible to indent it so that the sequence and nesting of the if/else can be easier to see. Mind doing that, please?

platform/qemu/virt/config.mk
```diff
@@ -30,12 +21,18 @@ PLATFORM_SYS_CLINT=y
 # Blobs to build
 FW_TEXT_START=0x80000000
 FW_JUMP=y
-# This needs to be 4MB alligned for 32-bit support
```
olofj: You're keeping the alignment but removing the comment. Even better would be a comment as to why 32-bit has different alignment requirements, and why we don't just use the same alignment on 64-bit too.

avpatel: I removed the comment because we have detailed comments in one place at platform/template/config.mk. If you still insist I will retain the comment here.

The alignment requirement comes from the Linux RISC-V MMU bootstrap logic requirements to work correctly on 32-bit and 64-bit systems. Unfortunately, these requirements are still undocumented in Linux RISC-V.

Makefile
avpatel: My bad for not indenting this properly. I will update this.

I could have certainly done "PLATFORM_RISCV_XLEN = $(OPENSBI_CC_XLEN)" here but these makes output of (i.e. "V=1" command-line parameter) less readable so I used a if-else-ladder to ensure that verbose make output is still readable.

avpatel:
Hi Olof, I have addressed all your comments. Please have a look again. --Anup

Issue #38, opened 2019-01-24
Unable to detect MicroSemi PCIe root complex when booting with openSBI on HiFive Unleased


alistair23:
As the device tree is not edited by the boot loader Linux doesn't see the MicroSemi PCIe controller.

alistair23:
Fixed with https://github.com/riscv/opensbi/pull/55

Issue #35, opened 2019-01-24
Unable to handle kernel NULL pointer dereference when snd_mtpav enabled


alistair23:
When CONFIG_SND_MTPAV is enabled in the kernel (5.0-rc3 + handful of required patches) I see this kernel NULL pointer dereference crash when booting using openSBI direct boot:
> [    4.282241] Unable to handle kernel NULL pointer dereference at virtual address 0000000000000379
> [    4.290786] Oops [#1]
> [    4.293019] Modules linked in:
> [    4.296064] CPU: 0 PID: 1 Comm: swapper/0 Not tainted 5.0.0-rc3-riscv #1
> [    4.302747] sepc: ffffffe0006d62a4 ra : ffffffe0006d689e sp : ffffffe1f7ac9a90
> [    4.309953]  gp : ffffffe000ae9dc8 tp : ffffffe1f7ac0000 t0 : ffffffe1f6e1fc00
> [    4.317158]  t1 : ffffffe0007bc500 t2 : 000000000000f9ac s0 : ffffffe1f7ac9ac0
> [    4.324365]  s1 : ffffffe1f7c56638 a0 : ffffffe1f7c56638 a1 : 00000000000000f5
> [    4.331570]  a2 : 0000000000000038 a3 : ffffffe1f7ac9957 a4 : ffffffe1f7c56063
> [    4.338776]  a5 : 0000000000000379 a6 : 0000000000000030 a7 : 0000000000000003
> [    4.345983]  s2 : ffffffe1f7c56638 s3 : 0000000000000000 s4 : 00000000000000f5
> [    4.353189]  s5 : 0000000000000008 s6 : ffffffe000ae9f4c s7 : ffffffe1f6e4d800
> [    4.360396]  s8 : ffffffe000000266 s9 : ffffffe000aea240 s10: ffffffe000aea230
> [    4.367601]  s11: 0000000000000000 t3 : 000000000000000f t4 : ffffffe000a95470
> [    4.374807]  t5 : 0000000000000040 t6 : ffffffe1f7c56060
> [    4.380104] sstatus: 0000000200000120 sbadaddr: 0000000000000379 scause: 000000000000000d
> [    4.388319] ---[ end trace ad20f5585df9d44a ]---
> [    4.392874] Kernel panic - not syncing: Attempted to kill init! exitcode=0x0000000b
> [    4.400513] ---[ end Kernel panic - not syncing: Attempted to kill init! exitcode=0x0000000b ]---

[...]

PR #55, opened 2019-02-12
Add support for specifying device tree for the Microsemi Expansion board

alistair23:
This adds support for the Microsemi expansion board. There seems to be a bug with how the DTB is copied that requires the third commit to work around.

### DIFF
--- a/platform/sifive/fu540/platform.c
+++ b/platform/sifive/fu540/platform.c
@@ -80,6 +80,7 @@ static void fu540_modify_dt(void *fdt)
 
 	plic_fdt_fixup(fdt, "riscv,plic0");
 }
+
 static int fu540_final_init(bool cold_boot)
 {
 	void *fdt;
--- a/docs/platform/sifive_fu540.md
+++ b/docs/platform/sifive_fu540.md
@@ -100,3 +100,13 @@ dd if=build/platform/sifive/fu540/firmware/fw_payload.bin of=/dev/disk2s1 bs=102
 ```
 bootm 0x80600000 - 0x82200000
 ```
+
+Booting SiFive Fu540 Platform with Microsemi Expansion board
+------------------------------------------------------------
+
+Until the Linux kernel has in-tree support for device trees and mainline u-boot is fully supported on the HiFive Unleashed you can follow these steps to boot Linux with the Microsemi expansion board. This method should not be copied on future boards and is considered a temporary solution until we can use a more standardised boot flow.
+
+To boot the Linux kernel with a device tree that has support for the Microsemi Expansion board you can include the following line when compiling the firmware:
+```
+FW_PAYLOAD_FDT="HiFiveUnleashed-MicroSemi-Expansion.dtb"
+```
--- a/firmware/fw_payload.S
+++ b/firmware/fw_payload.S
@@ -9,7 +9,7 @@
 
 #include "fw_base.S"
 
-	.align 3
+	.align 4
 	.section .entry, "ax", %progbits
 	.global fw_prev_arg1
 fw_prev_arg1:
@@ -21,7 +21,7 @@ fw_prev_arg1:
 #endif
 	ret
 
-	.align 3
+	.align 4
 	.section .entry, "ax", %progbits
 	.global fw_next_arg1
 fw_next_arg1:
@@ -33,7 +33,7 @@ fw_next_arg1:
 #endif
 	ret
 
-	.align 3
+	.align 4
 	.section .entry, "ax", %progbits
 	.global fw_next_addr
 fw_next_addr:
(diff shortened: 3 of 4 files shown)


alistair23:
This addresses https://github.com/riscv/opensbi/issues/38 and probably https://github.com/riscv/opensbi/issues/35 as well

olofj:
The slope is deep and slippery; do we really want opensbi to contain device trees for hardware support?

alistair23:
This series really only adds the device tree, but I do see your point.

I think ideally we shouldn't, but currently for the HiFive Unleashed we don't really have a better option unfortunately. Once we have SMP boot with mainline u-boot (or at least mainline u-boot working) we can remove it, but until then I don't see much of an option.

I can add some comments saying that new boards shouldn't do this if that helps?

include/sbi/sbi_scratch.h
```diff
@@ -57,6 +58,7 @@ struct sbi_scratch {
 	unsigned long hartid_to_scratch;
 	/** IPI type (or flags) */
 	unsigned long ipi_type;
+	unsigned long next_arg2;
 } __packed;
```
avpatel: Rename next_arg2 to prev_arg1. It is good to save Arg1 (a1 register) passed by previous booting stage in per-HART scratch space.

Also, define prev_arg1 before "next_arg1".

include/sbi/sbi_scratch.h
```diff
@@ -30,6 +30,7 @@
 #define SBI_SCRATCH_HARTID_TO_SCRATCH_OFFSET	(7 * __SIZEOF_POINTER__)
 /** Offset of ipi_type member in sbi_scratch */
 #define SBI_SCRATCH_IPI_TYPE_OFFSET		(8 * __SIZEOF_POINTER__)
+#define SBI_SCRATCH_NEXT_ARG2_OFFSET		(9 * __SIZEOF_POINTER__)
 /** Maximum size of sbi_scratch */
```
avpatel: Rename SBI_SCRATCH_NEXT_ARG2_OFFSET to SBI_SCRATCH_PREV_ARG1_OFFSET

platform/sifive/fu540/platform.c
```diff
@@ -53,11 +53,17 @@
 static void fu540_modify_dt(void *fdt)
 {
 	u32 i, size;
+	void *fdt_orig = ((void *)(sbi_scratch_thishart_ptr()->next_arg2));
 	int chosen_offset, err;
```
avpatel: Update as-per above comments.

firmware/fw_base.S
```diff
@@ -169,6 +169,8 @@ _start_warm:
 	/* Note: fw_next_arg1() uses a0, a1, and ra */
 	call	fw_next_arg1
 	REG_S	a0, SBI_SCRATCH_NEXT_ARG1_OFFSET(tp)
+	call	fw_prev_arg1
+	REG_S	a0, SBI_SCRATCH_NEXT_ARG2_OFFSET(tp)
 	/* Note: fw_next_addr() uses a0, a1, and ra */
```
avpatel: Save prev_arg1 before next_arg1.

include/sbi/sbi_scratch.h
avpatel: If we don't have built-in DTB linked to OpenSBI then prev_arg1 will point to "a1" register passed by previous booting stage.

platform/sifive/fu540/platform.c
```diff
@@ -69,10 +75,6 @@ static void fu540_modify_dt(void *fdt)
 			fdt_setprop_string(fdt, cpu_offset, "status", "masked");
 		memset(cpu_node, 0, sizeof(cpu_node));
 	}
-	size = fdt_totalsize(fdt);
-	err = fdt_open_into(fdt, fdt, size + 256);
-	if (err < 0)
-		sbi_printf("Device Tree can't be expanded to accmodate new node");
```
avpatel: Please add separate patch for changes in platform/sifive/fu540/platform.c

alistair23:
@avpatel Do you think this is the right direction? I thought that the fw_base.S should handle copying the FDT and that there is a bug in there which is why this is needed. If we go down this road it won't be generic and each platform will need to handle copying the FDT. My ideal solution would be to fix the problem in fw_base.S that doesn't copy the FDT, just need to figure out what that is.

atishp04:
I agree with Alistair. fdt_open_into will only work for platfroms that include libfdt. So we have to keep both fdt copying code in case we choose this route. We should fix it fw_base.S if possible.

avpatel:
Like Olof already mentioned, this is a slippery slop. I gave more thought for alternate approach. Here's what I think:

Instead of the updating "struct sbi_scratch" and fw_base.S, we can do the following in fu540_modify_dt():
1. "fdt" parameter already points to original DTB (passed by previous stage or built-in DTB) copied over to final location
2. Get size of DTB size using libfdt APIs
3. Copy over DTB from "fdt" to "orig_fdt" where orig_fdt = fdt + <dtb_size> + <increase_in_dtb_size>
4. Now use fdt_open_into() with "fdt" as new location of DTB and "orig_fdt" as old location of DTB

alistair23:
There seems to be a lot of confusion with what is going on.

I have pushed a v2 which fixes the original bug and removed the hack patch.

alistair23:
V3, just updated the docs to make it explicit that in the future this will be removed when no longer needed.

PR #56, opened 2019-02-13
csr fix

### LINKED ISSUE #49: Use CSR numbers when accessing CSR instead of CSR alias names
We should use CSR numbers everywhere to support older toolchains

atishp04:
Some older toolchain may not have all csr defined.
Update all csr functions to use defined CSR_ values instead of
toolchain defined values.

This fixes #49

### DIFF
--- a/firmware/fw_base.S
+++ b/firmware/fw_base.S
@@ -22,7 +22,7 @@ _start:
 	 * Jump to warm-boot if this is not the first core booting,
 	 * that is, for mhartid != 0
 	 */
-	csrr	a6, mhartid
+	csrr	a6, CSR_MHARTID
 	blt	zero, a6, _wait_for_boot_hart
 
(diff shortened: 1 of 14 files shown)


alistair23:
Can you tidy up the commit message? Something like this?
Some older toolchains may not have all csr's defined.
Update all the csr functions to use defined CSR_ values instead of
the toolchain defined values.

avpatel:
Can you please do same changes in firmware/ directory as well?

atishp04:
> Can you tidy up the commit message? Something like this?
> 
> > Some older toolchains may not have all the csr's defined.
> > Update all the csr functions to use the CSR_ #define values instead of
> > the toolchain defined values.

Done. Updated PR.

atishp04:
> Can you please do same changes in firmware/ directory as well?

Done. Updated PR.

PR #57, opened 2019-02-14
Makefile: Fix compilation on RV32

mickflemm:
On Makefile merge_objs calls ld without providing any hints on the target so on a multilib toolchain for example it will always pick the default one (elf64-littleriscv) which will then result a failure when compiling on RV32 since the objects will be 32bits and the target will be 64bits. The same happens on compile_ld that calls gcc without CFLAGS so it doesn't get mabi/march.

With this patch OpenSBI compiles on RV32 (PLATFORM_RISCV_XLEN=32), I tested fw_jump.elf on qemu (but it doesn't boot the kernel yet -bbl also doesn't boot the kernel so it may be something else).

Note that currently the newlib toolchain from riscv-gnu-toolchain doesn't support rv32imafdc (only rv32imafc) so since OpenSBI requires the d extension, it will not find the proper libgcc.a and default to the 64bit one, again resulting a failure. The glibc toolchain on the other hand works fine (and it doesn't really matter since we don't link with libc due to -nostdlib).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,8 @@ LDFLAGS		+=	$(platform-ldflags-y)
 LDFLAGS		+=	$(firmware-ldflags-y)
 
 MERGEFLAGS	+=	-r
+MERGEFLAGS	+=	-b elf$(PLATFORM_RISCV_XLEN)-littleriscv
+MERGEFLAGS	+=	-m elf$(PLATFORM_RISCV_XLEN)lriscv
 
 DTCFLAGS	=	-O dtb
 
@@ -225,7 +227,7 @@ compile_as = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     $(AS) $(ASFLAGS) $(call dynamic_flags,$(1),$(2)) -c $(2) -o $(1)
 compile_ld = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " LD        $(subst $(build_dir)/,,$(1))"; \
-	     $(CC) $(3) $(LDFLAGS) -Wl,-T$(2) -o $(1)
+	     $(CC) $(CFLAGS) $(3) $(LDFLAGS) -Wl,-T$(2) -o $(1)
 compile_ar = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " AR        $(subst $(build_dir)/,,$(1))"; \
 	     $(AR) $(ARFLAGS) $(1) $(2)


Makefile
```diff
@@ -171,6 +171,8 @@ LDFLAGS		+=	$(platform-ldflags-y)
 LDFLAGS		+=	$(firmware-ldflags-y)
 
 MERGEFLAGS	+=	-r
+MERGEFLAGS	+=	-b elf$(PLATFORM_RISCV_XLEN)-littleriscv
+MERGEFLAGS	+=	-m elf$(PLATFORM_RISCV_XLEN)lriscv
```
avpatel: Small correction ...

emulation mode will be "elf$(PLATFORM_RISCV_XLEN)lriscv_$(PLATFORM_RISCV_ABI)"

Makefile
```diff
@@ -225,7 +227,7 @@ compile_as = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     $(AS) $(ASFLAGS) $(call dynamic_flags,$(1),$(2)) -c $(2) -o $(1)
 compile_ld = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " LD        $(subst $(build_dir)/,,$(1))"; \
-	     $(CC) $(3) $(LDFLAGS) -Wl,-T$(2) -o $(1)
+	     $(CC) $(CFLAGS) $(3) $(LDFLAGS) -Wl,-T$(2) -o $(1)
 compile_ar = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
```
avpatel: Instead of using $(CFLAGS) for creating ELFs, we can add following flags to LDFLAGS:

LDFLAGS		+=	-Wl,-belf$(PLATFORM_RISCV_XLEN)-littleriscv
LDFLAGS		+=	-Wl,-melf$(PLATFORM_RISCV_XLEN)lriscv_$(PLATFORM_RISCV_ABI)

Makefile
mickflemm: You'll notice that this convention doesn't exist on newlib:
```riscv64-unknown-elf-ld: supported emulations: elf64lriscv elf32lriscv```

Also lp64d/ilp32d are not used, the default elf64lriscv / elf32lriscv implies the d extension, for more check out.
https://github.com/bminor/binutils-gdb/blob/master/ld/emulparams/elf32lriscv.sh
https://github.com/bminor/binutils-gdb/blob/master/ld/emulparams/elf64lriscv.sh
So for example if PLATFORM_RISCV_ABI is lp64d your approach will fail since there is no elf64lriscv_lp64d emulation.
```riscv64-unknown-linux-gnu-ld: supported emulations: elf64lriscv elf64lriscv_lp64f elf64lriscv_lp64 elf32lriscv elf32lriscv_ilp32f elf32lriscv_ilp32```

I think staying with the default one that implies the d extension (since it's needed anyway for OpenSBI) is more portable and cleaner, at least for now. If OpenSBI supports builds without the d extension then we could add more logic there to properly handle emulation.

Makefile
mickflemm: Since you are using CC instead of LD directly, I think the march / mabi approach is more appropriate (plus there is the issue above regarding lp64d/ilp32d that will complicate things for no reason).

avpatel: Makes sense but this means only "-mabi", "-march", and "-mcmodel" are required in $(LDFLAGS). Better we append only these options to $(LDFLAGS) instead of all options in $(CFLAGS).

Makefile
avpatel: Sure, no problem.

Let's use "elf$(PLATFORM_RISCV_XLEN)lriscv" as emulation mode.

avpatel:
To update tile and description of a commit in GitHub PR, you can do the following:
1. First extract your commit as patch
2. Roll-back your HEAD to previous commit
3. Apply the patch
4. Commit the changes with new title and description
5. Do "git push origin master --force"

Makefile
mickflemm: -mabi is not understood by ld though, it'll be misleading to have it on LDFLAGS, it's not a big deal to pass CFLAGS to gcc, it's cleaner this way and more consistent IMHO.

mickflemm:
I re-named the title to be consistent with your guidelines, however editing the description is a bit of a mess. I'll try it out and see if it works...

Makefile
avpatel: No problem, we can fix compile_ld() function in a different patch. Let's go with your change.

One last thing, first line of commit description is treated as subject of commit. This is "Make sure ld is always aware of the target abi/emulation" right now. Please update it to "Makefile: Make sure ld is always aware of the target abi/emulation" so that it is consistent with our commit description style. Same commit description style is followed in Linux and U-Boot as well.

My apologies for being picky about commit description style. We are trying to follow Linux coding style here so that developers don't have to adapt to new coding style.

avpatel:
Created GitHub issue for compile_ld() fix.

Refer,
https://github.com/riscv/opensbi/issues/58

Makefile
mickflemm: ACK I hope it's ok now ;-)

PR #59, opened 2019-02-15
Add 'make run' command for platform specific run

avpatel:
Makes for easy and quick build-run one-stop command.

For now only added for qemu targets. It can be added for
any platform having simulator/emulator (such as QEMU).

Signed-off-by: Anup Patel <anup.patel@wdc.com>


alistair23:
Can you add the -nographic option to the QEMU args?

avpatel:
Done. Replaced "-display none -serial stdio" with "-nographic".

PR #60, opened 2019-02-15
QEMU-Virt: Implement system shutdown

mickflemm:
In order for QEMU to be compatible with Spike, it implements a simple protocol used for reporting back the simulation's status, through the memory-mapped "test finisher" device. We use that protocol to make QEMU exit on system shutdown.

In case the above fails, we fall back to sending an IPI_EVENT_HALT to every hart on the system.

### DIFF
--- a/platform/qemu/virt/platform.c
+++ b/platform/qemu/virt/platform.c
@@ -5,11 +5,14 @@
  *
  * Authors:
  *   Anup Patel <anup.patel@wdc.com>
+ *   Nick Kossifidis <mick@ics.forth.gr>
  */
 
 #include <sbi/riscv_encoding.h>
+#include <sbi/riscv_io.h>
 #include <sbi/sbi_const.h>
 #include <sbi/sbi_hart.h>
+#include <sbi/sbi_ipi.h>
 #include <sbi/sbi_platform.h>
 #include <plat/irqchip/plic.h>
 #include <plat/serial/uart8250.h>
@@ -19,6 +22,8 @@
 #define VIRT_HART_STACK_SIZE		8192
 
 #define VIRT_TEST_ADDR			0x100000
+#define VIRT_TEST_FINISHER_FAIL		0x3333
+#define VIRT_TEST_FINISHER_PASS		0x5555
 
 #define VIRT_CLINT_ADDR			0x2000000
 
@@ -122,8 +127,29 @@ static int virt_timer_init(bool cold_boot)
 
 static int virt_system_down(u32 type)
 {
-	/* For now nothing to do. */
-	return 0;
+	struct sbi_scratch *local_scratch = NULL;
+	int ret = 0;
+
+	/* Tell the "finisher" that the simulation
+	 * was successful so that QEMU exits
+	 */
+	writew(VIRT_TEST_FINISHER_PASS, (void *)VIRT_TEST_ADDR);
+
+	/* The above didn't work, send an IPI on
+	 * everyone to hang
+	 */
+	local_scratch = sbi_scratch_thishart_ptr();
+
+	ret = sbi_ipi_send_many(local_scratch,
+				NULL, SBI_IPI_EVENT_HALT);
+
+	/* We should be done by now, if not something
+	 * weird happened, all we can do is hang
+	 */
+	sbi_hart_hang();
+
+	/* Make the compiler happy */
+	return ret;
 }
 
 struct sbi_platform platform = {
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,8 @@ LDFLAGS		+=	$(platform-ldflags-y)
 LDFLAGS		+=	$(firmware-ldflags-y)
 
 MERGEFLAGS	+=	-r
+MERGEFLAGS	+=	-b elf$(PLATFORM_RISCV_XLEN)-littleriscv
+MERGEFLAGS	+=	-m elf$(PLATFORM_RISCV_XLEN)lriscv
 
 DTCFLAGS	=	-O dtb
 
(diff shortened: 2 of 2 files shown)


platform/qemu/virt/platform.c
```diff
@@ -122,8 +127,29 @@ static int virt_timer_init(bool cold_boot)
...
+	writew(VIRT_TEST_FINISHER_PASS, (void *)VIRT_TEST_ADDR);
+
+	/* The above didn't work, send an IPI on
+	 * everyone to hang
+	 */
+	local_scratch = sbi_scratch_thishart_ptr();
+
+	ret = sbi_ipi_send_many(local_scratch,
+				NULL, SBI_IPI_EVENT_HALT);
+
```
avpatel: It's good to send halt event in-case platform specific shutdown failed.

I think the sbi_ipi_send_many() call should called from generic library so that we have it for all platforms. You can add this to sbi_system_reboot() and sbi_system_shutdown() functions of lib/sbi_shutdown.c. You can use "lib: " prefix for changes in generic library.

platform/qemu/virt/platform.c
```diff
@@ -122,8 +127,29 @@ static int virt_timer_init(bool cold_boot)
...
-	/* For now nothing to do. */
-	return 0;
+	struct sbi_scratch *local_scratch = NULL;
+	int ret = 0;
+
+	/* Tell the "finisher" that the simulation
+	 * was successful so that QEMU exits
+	 */
+	writew(VIRT_TEST_FINISHER_PASS, (void *)VIRT_TEST_ADDR);
+
```
avpatel: The virt_system_down() should only have writew(). Other things can be move to generic library. See below comment.

platform/qemu/virt/platform.c
```diff
@@ -122,8 +127,29 @@ static int virt_timer_init(bool cold_boot)
...
+	local_scratch = sbi_scratch_thishart_ptr();
+
+	ret = sbi_ipi_send_many(local_scratch,
+				NULL, SBI_IPI_EVENT_HALT);
+
+	/* We should be done by now, if not something
+	 * weird happened, all we can do is hang
+	 */
+	sbi_hart_hang();
+
```
avpatel: We already have sbi_hart_hang() in sbi_system_reboot() and sbi_system_shutdown() so this is not required here.

avpatel:
Nick,

Something is not write with this PR because I see a commit which is already merged in riscv/opensbi.

I realize that Github PR system is quite different from the usual e-mail based review system but I will try my best to help you.

General approach when creating a new GitHub PR is as follows:
1. Create a fork of main repo (You already have this)
2. Add main repo URL as "upstream" using "git remote add upstream https://github.com/riscv/opensbi.git"
3. Keep the "master" branch of your fork in-sync with "master" branch of main repo using "git reset --hard upstream/master" (I think you might have used "git merge" here that's we see a dummy merge commit in your PR)
3. To work on something new, create <abc> branch in based on your "master" branch.
4. Add commits to your "<abc>" branch and push it to your fork
5. Create Github PR based on "<abc>" branch in your fork

When we want to revise commits in "<abc>" branch of your fork do the following:
1. Extract patches from your "<abc>" branch using "git format-patch -M -C origin/master"
2. Roll-back "<abc>" branch to your "master" using "git reset --hard origin/master"
3. Apply patches again with comments addressed to "<abc>" branch
4. Force push "<abc>" branch using "git push origin <abc> --force"

Regards,
Anup

avpatel:
You can use "platform: " prefix for changes in platform/directory so
"QEMU-Virt: Implement system shutdown"
can be
"platform: virt: Implement system shutdown"

These commit subject prefixes are described in docs/contributing.md

mickflemm:
I'll close this PR and submit a new one

platform/qemu/virt/platform.c
mickflemm: I get it for sbi_system_shutdown but why on reboot ?

platform/qemu/virt/platform.c
mickflemm: ACK makes sense

platform/qemu/virt/platform.c
mickflemm: OK I'll make a new PR

platform/qemu/virt/platform.c
avpatel: In both shutdown and reboot, if platform specific method fails to do the job then we should atleast halt other HARTs. Do you agree ?

mickflemm: I'm ok if we halt all other harts (vs everyone) since the current hart will then be able to do the reboot, but halting everyone or everyone else and then the current hart doesn't make sense for reboot (this way reboot == halt). We can simply return an error and let the OS handle it (e.g. on Linux we could trigger an architecture-independent restart method, check out https://elixir.bootlin.com/linux/latest/source/include/linux/reboot.h#L64).

Also reboot != reset, we need to have a platform-specific reset method where all other harts are hanged and the current one triggers a reset through a power domain controller/pmic/gpio pin/whatever and if that fails it'll fall-back to reboot. To me reboot means "clean up and execute the second stage boot loader" (e.g. u-boot), or in case of the kernel "clean up and re-execute the kernel". Reset on the other hand is "power-cycle the system".

That means in case of re-boot we shouldn't use sbi_hart_hang on the other harts because since the while loop is there it will never recover. I think we should have an sbi_hart_wait function where we disable external interrupts (so that another hart can't get an interrupt from a device and try to execute code driver code during reboot) and mark the hart unavailable (so if the OS sends an IPI during reboot it'll get ignored), which basically means wfi where interrupt == IPI (IPI will get un-blocked after re-boot where all harts will be marked as available again).

What do you think ?

avpatel: I think we are on same page here.

Currently, we don't have a clear definition of Reboot SBI call in SBI spec so our reboot implementation is crude.

I think we should wait for SBI PM extension to do changes in sbi_system_reboot() implementation so let's leave it unmodified now.

PR #63, opened 2019-02-17
System shutdown updates

mickflemm:
This series provides a platform-specific system shutdown implementation for QEMU/Virt and a generic implementation that also handles SMP systems as fall-back, or for platforms that don't have a system shutdown implementation yet.

### DIFF
--- a/lib/sbi_ipi.c
+++ b/lib/sbi_ipi.c
@@ -5,6 +5,7 @@
  *
  * Authors:
  *   Anup Patel <anup.patel@wdc.com>
+ *   Nick Kossifidis <mick@ics.forth.gr
  */
 
 #include <sbi/riscv_asm.h>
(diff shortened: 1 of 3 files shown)


lib/sbi_ipi.c
```diff
@@ -5,6 +5,7 @@
  *
  * Authors:
  *   Anup Patel <anup.patel@wdc.com>
+ *   Nick Kossifidis <mick@ics.forth.gr
```
avpatel: Small typo here. You forgot a ">" here.

mickflemm: Oops, missed that one, I just pushed a fix.

avpatel:
For this PR it is fine but you don't need to add separate commit for addressing comments.

You can just force update your "local_fixes" branch with updated commits.

PR #64, opened 2019-02-18
Fix error evaluating OPENSBI_CC_XLEN on Ubuntu

avpatel:
We get following error evaluating OPENSBI_CC_XLEN on
Ubuntu-18.04:
/bin/sh: 1: Bad substitution

This PR fixes above error by using "awk"


avpatel:
Done. Replaced use of awk with sed.

PR #68, opened 2019-02-19
Handle M-mode to M-mode traps without overwriting stack

avpatel:


atishp04:
Looks good to me. 

I observed that 
Single line comment format in header file is different from the assembly files. Is that intentional ?

avpatel:
When we document a member of structure OR some function we have to use doxygen c-style comments which is "/** <description */

The inline comments in assembly describe code and not any structure member or function so that's why these are regular c-style comments "/* <description> */"

BTW, currently doxygen will not parse assembly files.

avpatel:
Tested on QEMU virt machine as well as SiFive unleashed.

Merging this now.

PR #73, opened 2019-02-22
Makefile: Set the platform variables before parsing the platforms

alistair23:
Ensure the platform variable PLATFORM_RISCV_XLEN is set before we parse
the platform files.

This fixes the 32-bit openSBI FW_JUMP_ADDR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,15 @@ DTC		=	dtc
 # Guess the compillers xlen
 OPENSBI_CC_XLEN := $(shell TMP=`$(CC) -dumpmachine | sed 's/riscv\([0-9][0-9]\).*/\1/'`; echo $${TMP})
 
+# Setup platform XLEN, ABI, ISA and Code Model
+ifndef PLATFORM_RISCV_XLEN
+  ifeq ($(OPENSBI_CC_XLEN), 32)
+    PLATFORM_RISCV_XLEN = 32
+  else
+    PLATFORM_RISCV_XLEN = 64
+  endif
+endif
+
 # Setup list of objects.mk files
 ifdef PLATFORM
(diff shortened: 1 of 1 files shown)


avpatel:
I had intentionally kept this after including config.mk to provide platform support code to explicitly set PLATFORM_RISCV_XLEN.

I guess pre-setting PLATFORM_RISCV_XLEN before including config.mk would also work because Makefile allows multiple assignment to same variable.

avpatel:
I just checked.

This patch breaks the case where we want to set PLATFORM_RISCV_XLEN in platform config.mk because PLATFORM_RISCV_ABI and PLATFORM_RISCV_ISA are depend on PLATFORM_RISCV_XLEN.

I have a slightly better suggestion ....

Do following after setting OPENSBI_CC_XLEN:

# Setup platform XLEN
ifndef PLATFORM_RISCV_XLEN
  ifeq ($(OPENSBI_CC_XLEN), 32)
    PLATFORM_RISCV_XLEN = 32
  else
    PLATFORM_RISCV_XLEN = 64
  endif
endif

Do following just before "Setup compilation commands flags":

# Setup platform ABI, ISA and Code Model
ifndef PLATFORM_RISCV_ABI
  ifeq ($(PLATFORM_RISCV_XLEN), 32)
    PLATFORM_RISCV_ABI = ilp$(PLATFORM_RISCV_XLEN)
  else
    PLATFORM_RISCV_ABI = lp$(PLATFORM_RISCV_XLEN)
  endif
endif
ifndef PLATFORM_RISCV_ISA
  PLATFORM_RISCV_ISA = rv$(PLATFORM_RISCV_XLEN)imafdc
endif
ifndef PLATFORM_RISCV_CODE_MODEL
  PLATFORM_RISCV_CODE_MODEL = medany
endif

alistair23:
V2 sent with the requested changes

PR #74, opened 2019-02-27
Fix makefile problems

andreas-schwab:


Makefile
```diff
@@ -325,7 +325,7 @@ ifneq ($(platform-runcmd),)
 	$(platform-runcmd) $(RUN_ARGS)
 else
 ifdef PLATFORM
-	@echo Platform $(PLATFORM) doesn't specify a run command
+	@echo "Platform $(PLATFORM) doesn't specify a run command"
 	@false
 else
 	@echo Run command only available when targeting a platform
```
damien-lemoal: You probably need quotes here too.

andreas-schwab: No.

Makefile
```diff
@@ -213,15 +213,15 @@ compile_cpp = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     $(CPP) $(CPPFLAGS) -x c $(2) | grep -v "\#" > $(1)
 compile_cc_dep = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " CC-DEP    $(subst $(build_dir)/,,$(1))"; \
-	     echo `dirname $(1)`/ \\  > $(1) && \
+	     printf %s `dirname $(1)`/  > $(1) && \
```
damien-lemoal: More on quotes: does the `dirname $(1)` need to be double quoted here ? Can there be a problem if a path component name has a space ?

andreas-schwab: Then the whole makefile collapses.

damien-lemoal: Fair enough

PR #75, opened 2019-02-28
Warm Reset fix

atishp04:


firmware/fw_base.S
```diff
@@ -189,6 +189,10 @@ _start_warm:
 	/* Setup trap handler */
 	la	a4, _trap_handler
 	csrw	CSR_MTVEC, a4
+1:
+	csrr	a5, CSR_MTVEC
+	bne	a4, a5, 1b
+
```
avpatel: Unwanted empty line after "bne". Otherwise looks good.

firmware/fw_base.S
```diff
@@ -124,6 +126,10 @@ _wait_for_boot_hart:
 	REG_L	a5, (a4)
 	beqz	a5, _wait_for_boot_hart
 
+	csrr	a6, CSR_MHARTID
+	beqz	a6, _start_warm
+	li	ra, 0
+	call	_reset_regs
```
avpatel: I think you should unconditionally reset all registers in _start_warm() for all HARTs.

atishp04: ok. Updated and created another PR #80

PR #76, opened 2019-02-28
Serial fix

atishp04:


xerpi:
I think it looks good overall, but shouldn't the functions that return the character (`sifive_uart_getc` and `uart8250_getc`) return `-1` instead of `0` when there isn't a pending character?

atishp04:
> I think second patch should be squashed with this patch because applying just this patch will cause compile error.
> 
> We should generally try to ensure that each individual patch in a patchset don't break compilation. This helps in future when doing git bisect.

Sure. Done.

atishp04:
> I think it looks good overall, but shouldn't the functions that return the character (`sifive_uart_getc` and `uart8250_getc`) return `-1` instead of `0` when there isn't a pending character?

 -1 should be for failure. When there isn't a pending character, the fifo is just empty. As per my understanding, that's different from a failure. @avpatel Any thoughts ?

avpatel:
We have few error codes defined in include/sbi/sbi_error.h

May be you can add SBI_ENOTAVAIL to include/sbi/sbi_error.h and return SBI_ENOTAVAIL instead of -1

xerpi:
I think that would work since Linux expects it to be positive, so any negative number will be treated as no more pending characters available.

By the way, I have tried booting Linux returning `0` when there's no character available and the Linux SBI console driver thinks it's [returning valid characters](https://github.com/torvalds/linux/blob/master/drivers/tty/hvc/hvc_riscv_sbi.c#L33).

atishp04:
updated the PR.

PR #78, opened 2019-03-04
firmware: Move _boot_hart_done to the data section & Add a barrier instruction for wait for boot hart

wxjstz:
Writable code section can cause some security problems, so move _boot_hart_done
to the data section

Signed-off-by: Xiang Wang <wxjstz@126.com>

### DIFF
--- a/firmware/fw_base.S
+++ b/firmware/fw_base.S
@@ -114,6 +114,7 @@ _fdt_reloc_done:
 
 	/* Wait for boot hart */
 _wait_for_boot_hart:
+	fence	rw, rw
 	la	a4, _boot_hart_done
 	REG_L	a5, (a4)
 	beqz	a5, _wait_for_boot_hart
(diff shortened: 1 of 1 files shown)


firmware/fw_base.S
```diff
@@ -13,6 +13,11 @@
 #include <sbi/sbi_scratch.h>
 #include <sbi/sbi_trap.h>
 
+	.align 3
+	.section .data, "aw"
+_boot_hart_done:
+	RISCV_PTR	0
+
```
avpatel: It's good to have _boot_hart_done in data section but let's try to keep _start as first symbol in fw_base.S for readability.

I would suggest to have _boot_hart_done just before _hartid_to_scratch.

wxjstz: done

wxjstz:
> /s/instructionto/instructions to

Done

PR #80, opened 2019-03-04
Firmware fixes

atishp04:


firmware/fw_base.S
```diff
@@ -18,6 +18,57 @@
...
+	REG_S	a4, SBI_SCRATCH_NEXT_MODE_OFFSET(tp)
+	la	a4, _start_warm
+	REG_S	a4, SBI_SCRATCH_WARMBOOT_ADDR_OFFSET(tp)
+	la	a4, platform
+	REG_S	a4, SBI_SCRATCH_PLATFORM_ADDR_OFFSET(tp)
+	la	a4, _hartid_to_scratch
+	REG_S	a4, SBI_SCRATCH_HARTID_TO_SCRATCH_OFFSET(tp)
+	REG_S	zero, SBI_SCRATCH_IPI_TYPE_OFFSET(tp)
+	REG_S	zero, SBI_SCRATCH_TMP0_OFFSET(tp)
+
```
avpatel: This will break in the case when SOC provides HW mechanism to bringup HART's because in this situation only one HART will boot from _start and reset all HART will start from _start_warm.

We have to setup scratch space for all HARTs from boot HART using an assembly loop.

firmware/fw_base.S
```diff
@@ -175,6 +175,13 @@ _wait_for_boot_hart:
...
 
+	li	ra, 0
+	call	_reset_regs
+
+	csrr	tp, CSR_MSCRATCH
+	/* Setup stack */
+	add	sp, tp,zero
+
 _start_warm:
 	/* Disable and clear all interrupts */
```
avpatel: I think you missed my last comment.

We should call _reset_regs() in _start_warm() so that its done for all HART. This will also work for HART hotplug.

firmware/fw_base.S
atishp04: Ahh yes. I will create a separate patch for this and change accordingly.

firmware/fw_base.S
atishp04: argh. Yes. _reset_regs call was supposed to be after _start_warm. I think I messed up while switching to latest branch. Sorry.

atishp04:
Updated the PR.

Issue #65, opened 2019-02-18
sbi_trap_error: hart2: trap handler failed (error -5)


andreas-schwab:
sbi_trap_error: hart2: trap handler failed (error -5)
sbi_trap_error: hart2: mcause=0xd mtval=0x155574e000
sbi_trap_error: hart2: mepc=0x80003dba mstatus=0x8000000a00027822
sbi_trap_error: hart2: ra=0x80001eaa sp=0x80011d60
sbi_trap_error: hart2: gp=0x2aaaccdd18 tp=0x155573cb30
sbi_trap_error: hart2: s0=0x80011dc0 s1=0x5e
sbi_trap_error: hart2: a0=0x20000 a1=0x80011df0
sbi_trap_error: hart2: a2=0x15555eee96 a3=0x8000000a00006022
sbi_trap_error: hart2: a4=0x80011d6b a5=0x155574e001
sbi_trap_error: hart2: a6=0x0 a7=0x155574e001
sbi_trap_error: hart2: s2=0xee s3=0x3
sbi_trap_error: hart2: s4=0x96 s5=0x20
sbi_trap_error: hart2: s6=0x0 s7=0x15555eee96
sbi_trap_error: hart2: s8=0x74 s9=0x14774
sbi_trap_error: hart2: s10=0x58782b5b s11=0x34fe3e52
sbi_trap_error: hart2: t0=0x4 t1=0x155574dffd
sbi_trap_error: hart2: t2=0x55 t3=0x85a303
sbi_trap_error: hart2: t4=0x0 t5=0x0
sbi_trap_error: hart2: t6=0x15

The kernel was booted via u-boot.

atishp04:
Do you have the U-Boot SMP patches included ?
Without that, you should provide FU540_ENABLED_HART_MASK=0x02 as an compile time option.

Refer: https://github.com/riscv/opensbi/blob/master/docs/platform/sifive_fu540.md

andreas-schwab:
Yes, of course.

andreas-schwab:
I'm also often seeing trap errors while opensbi is still booting.

sbi_trap_error: hart1: misaligned store handler failed (error -10)

OpenSBI v0.1 (Feb 14 2019 12:10:27)
sbi_trap_error: hart1: mcause=0x6 mtval=0xfffffffffffffff9
   ____                    _____ ____ _____
  / __ \                  / ____|  _ \_   _|
 | |  | |_ __   ___ _ __ | (___ | |_) || |
[...]

PR #83, opened 2019-03-06
Fence revert

atishp04:


wxjstz:
warm reset issue ???

avpatel:
Yes, some thing out-of-sync when we use "reset button" (warm boot) on SiFive Unleashed board. Although, cold-boot using power-on button works perfectly fine on SiFive Unleashed board.

Refer, https://github.com/riscv/opensbi/issues/65

We have few fixes in-place for warm boot (submitted by Atish) but the "fence rw,rw" patch is again causing warm boot issue to re-appear so for time-being we have reverted it.

atishp04:
No need to merge this. I have raised another PR with fence revert and some additional required warm reset fix.

alistair23:
Closing the PR as requested

PR #84, opened 2019-03-06
Warm reset fix

atishp04:
This PR contains the 2nd part of warm reset fix. It also includes the changes in previous PR (#83 ).

### DIFF
--- a/firmware/fw_base.S
+++ b/firmware/fw_base.S
@@ -114,7 +114,6 @@ _fdt_reloc_done:
 
 	/* Wait for boot hart */
 _wait_for_boot_hart:
-	fence	rw, rw
 	la	a4, _boot_hart_done
 	REG_L	a5, (a4)
 	beqz	a5, _wait_for_boot_hart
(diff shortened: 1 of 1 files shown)


firmware/fw_base.S
```diff
@@ -116,6 +116,9 @@ _fdt_reloc_done:
 _wait_for_boot_hart:
 	la	a4, _boot_hart_done
 	REG_L	a5, (a4)
+	nop
```
alistair23: Can you please add a comment inline to describe why this is required?

Also, why can't this be a fence? That seems like a bug somewhere else if a fence doesn't work

atishp04: > Can you please add a comment inline to describe why this is required?
> 
Done.
> Also, why can't this be a fence? That seems like a bug somewhere else if a fence doesn't work

One possible explanation is non-boot cpu cache may have some stale corrupt DTB data which gets written. But this is a just theory and no way to prove it.

That's why commit text says that it's being reverted just for the time being until we figure out the actual reason.

Issue #81, opened 2019-03-06
Fails to compile with gcc 9.0


kraj:
see https://errors.yoctoproject.org/Errors/Details/231564/

alistair23:
Interesting, GCC9 is pointing out something which is a bad idea to have in code. For RISC-V it shouldn't cause catastrophic problems, but will slow things down. Even then our struct should always be aligned, so this really won't cause us any issues for us.

For more details you can read the GCC bug report that started this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628

I'll work on a patch, I think an intermediate variable will fix it, as that way we never take an address of a packed struct. A little bit of a pain, but I don't see a much better solution.

alistair23:
@kraj Do you have a branch with GCC9 that I can use?

kraj:
Use Yoe Distro ( shameless plug ) which I co-maintain and use yoe/mut branch
https://github.com/YoeDistro/yoe-distro

or you can use yoe/mut branch from my contrib repo
https://github.com/YoeDistro/openembedded-core/tree/yoe/mut

and in `local.conf` set

```
GCCVERSION = "9.%"
```

kraj:
> Interesting, GCC9 is pointing out something which is a bad idea to have in code. For RISC-V it shouldn't cause catastrophic problems, but will slow things down. Even then our struct should always be aligned, so this really won't cause us any issues for us.
> 

I think we should avoid typecasting packed structs to pointers, or not pack them

> For more details you can read the GCC bug report that started this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628
> 

Yes I am aware of that

> I'll work on a patch, I think an intermediate variable will fix it, as that way we never take an address of a packed struct. A little bit of a pain, but I don't see a much better solution.

that will be awesome

PR #86, opened 2019-03-07
lib: Don't access address of packed struct member

alistair23:
Taking the address of a packed member may result in an unalligned
pointer access. This can cause slowdowns for RISC-V and breaks
compilation with GCC9.

To fix the problem let's just store the address into an alligned temp
variable before we access it. This seems the simplest solution as we
don't want to unpack the struct.

This fixes https://github.com/riscv/opensbi/issues/81.

### DIFF
--- a/lib/sbi_ipi.c
+++ b/lib/sbi_ipi.c
@@ -23,6 +23,7 @@ static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid, u32 event)
 {
 	struct sbi_scratch *remote_scratch = NULL;
 	const struct sbi_platform *plat = sbi_platform_ptr(scratch);
+	unsigned long tmp = remote_scratch->ipi_type;
 
 	if (sbi_platform_hart_disabled(plat, hartid))
 		return -1;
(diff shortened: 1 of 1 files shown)


lib/sbi_ipi.c
```diff
@@ -23,6 +23,7 @@ static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid, u32 event)
 {
 	struct sbi_scratch *remote_scratch = NULL;
 	const struct sbi_platform *plat = sbi_platform_ptr(scratch);
+	unsigned long tmp = remote_scratch->ipi_type;
```
avpatel: For sure, this breaks atomicity of "ipi_type". The "tmp" variable is on stack so it's location will change based on calling path.

We have three options here:
1. Drop __packed from "struct sbi_scratch" since all members are "unsigned long". May be some compiler expert can clarify if this would work fine.
2. Change "ipi_type" to "ipi_data" (pointer) which will point to per-HART "struct sbi_ipi_data" instance. The "struct sbi_ipi_data" will have "unsigned long ipi_type" member.
3. We have ensured that most packed structures have members with appropriate alignment so we suppress the GCC9 warning using some GCC9 option.

alistair23: Yep, you are right. This definitely does not work. Let me check this again tomorrow.

alistair23: Why don't we just change ipi_type to be a pointer to an unsigned long? It's never set in assembly only ever by the atomics.

avpatel:
Go ahead with option 2 because we want pass more information from one HART to another for REMOTE sfence.vma SBI calls.

PR #88, opened 2019-03-08
lib: Create a sbi_ipi_data structure

alistair23:
Create a sbi_ipi_data structure that holds unpacked IPI information. At
the same time remove ipi_type from the sbi_scratch struct and use a
fixed offset to access it.

This structure fits in behind the sbi_scratch structure.

This fixes https://github.com/riscv/opensbi/issues/81

### DIFF
--- a/include/sbi/sbi_scratch.h
+++ b/include/sbi/sbi_scratch.h
@@ -28,16 +28,23 @@
 #define SBI_SCRATCH_PLATFORM_ADDR_OFFSET	(6 * __SIZEOF_POINTER__)
 /** Offset of hartid_to_scratch member in sbi_scratch */
 #define SBI_SCRATCH_HARTID_TO_SCRATCH_OFFSET	(7 * __SIZEOF_POINTER__)
-/** Offset of ipi_type member in sbi_scratch */
-#define SBI_SCRATCH_IPI_TYPE_OFFSET		(8 * __SIZEOF_POINTER__)
 /** Offset of tmp0 member in sbi_scratch */
-#define SBI_SCRATCH_TMP0_OFFSET			(9 * __SIZEOF_POINTER__)
-/** Maximum size of sbi_scratch */
-#define SBI_SCRATCH_SIZE			256
+#define SBI_SCRATCH_TMP0_OFFSET			(8 * __SIZEOF_POINTER__)
+
+/** sbi_ipi_data is located behind sbi_scratch. This struct is not packed. */
+/** Offset of ipi_type in sbi_ipi_data */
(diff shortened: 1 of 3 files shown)


alistair23:
The new sbi_ipi_data struct is located in the middle of the reserved memory to try to avoid having to change offsets in the future. It seems like the easiest option to me while still avoiding too much future churn

include/sbi/sbi_scratch.h
```diff
@@ -28,17 +28,24 @@
...
-#define SBI_SCRATCH_IPI_TYPE_OFFSET		(8 * __SIZEOF_POINTER__)
 /** Offset of tmp0 member in sbi_scratch */
-#define SBI_SCRATCH_TMP0_OFFSET			(9 * __SIZEOF_POINTER__)
-/** Maximum size of sbi_scratch */
-#define SBI_SCRATCH_SIZE			256
+#define SBI_SCRATCH_TMP0_OFFSET			(8 * __SIZEOF_POINTER__)
+
+/** sbi_ipi_data is located behind sbi_scratch. This struct is not packed. */
+/** Offset of ipi_type in sbi_ipi_data */
+#define SBI_IPI_DATA_IPI_TYPE_OFFSET		(15 * __SIZEOF_POINTER__)
```
atishp04: Not sure we should keep it at a larger offset. It all depends whether we see more members are getting added to sbi_scratch in future or we will have more items outside the sbi_scratch.

By keeping at (15 * __SIZEOF_POINTER__), we are restricting sbi_scratch can grow to only up to that.
This should be fine I guess. Just food for thought ;)

include/sbi/sbi_scratch.h
```diff
@@ -28,17 +28,24 @@
...
+
+/** Maximum size of sbi_scratch and sbi_ipi_data */
+#define SBI_SCRATCH_SIZE			(32 * __SIZEOF_POINTER__)
```
atishp04: May be rename it to SBI_SCRATCH_TOTAL_SIZE ?

It is bit confusing as the structure has the same name.

include/sbi/sbi_scratch.h
alistair23: It's still possible to change it in the future, so we aren't locked in forever

avpatel:
Overall looks good to me.

I have a small suggestion. Instead on initialising ipi_data in fw_base.S you can initialize it in sbi_ipi_init().

alistair23:
Updated with the init being done in sbi_ipi_init()

alistair23:
I haven't changed the total size macro name, let me know if we think it should be changed. I'm happy as it is, but can change it

PR #89, opened 2019-03-09
Sfence fix

atishp04:
#88 is merged. It can be applied cleanly.

### DIFF
--- a/lib/sbi_ipi.c
+++ b/lib/sbi_ipi.c
@@ -19,10 +19,13 @@
 #include <sbi/sbi_timer.h>
 #include <sbi/sbi_unpriv.h>
 
-static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid, u32 event)
+static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid,
+			u32 event, void *data)
 {
 	struct sbi_scratch *remote_scratch = NULL;
 	const struct sbi_platform *plat = sbi_platform_ptr(scratch);
+	struct sbi_tlb_info *tlb_info = data;
+	struct sbi_tlb_info *ipi_tlb_data;
 
 	if (sbi_platform_hart_disabled(plat, hartid))
 		return -1;
@@ -31,6 +34,13 @@ static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid, u32 event)
 	 * trigger the interrupt
 	 */
 	remote_scratch = sbi_hart_id_to_scratch(scratch, hartid);
+	if (event == SBI_IPI_EVENT_SFENCE_VMA ||
+	    event == SBI_IPI_EVENT_SFENCE_VMA_ASID) {
+		ipi_tlb_data = sbi_tlb_info_ptr(remote_scratch);
+		ipi_tlb_data->start = tlb_info->start;
+		ipi_tlb_data->size = tlb_info->size;
+		ipi_tlb_data->asid = tlb_info->asid;
+	}
 	atomic_raw_set_bit(event, &sbi_ipi_data_ptr(remote_scratch)->ipi_type);
 	mb();
 	sbi_platform_ipi_send(plat, hartid);
@@ -41,7 +51,7 @@ static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid, u32 event)
 }
 
 int sbi_ipi_send_many(struct sbi_scratch *scratch,
-		      ulong *pmask, u32 event)
+		      ulong *pmask, u32 event, void *data)
 {
 	ulong i, m;
 	ulong mask = sbi_hart_available_mask();
@@ -53,13 +63,13 @@ int sbi_ipi_send_many(struct sbi_scratch *scratch,
 	/* send IPIs to every other hart on the set */
 	for (i = 0, m = mask; m; i++, m >>= 1)
 		if ((m & 1UL) && (i != hartid))
-			sbi_ipi_send(scratch, i, event);
+			sbi_ipi_send(scratch, i, event, data);
 
 	/* If the current hart is on the set, send an IPI
 	 * to it as well
 	 */
 	if (mask & (1UL << hartid))
-		sbi_ipi_send(scratch, hartid, event);
+		sbi_ipi_send(scratch, hartid, event, data);
 
 
 	return 0;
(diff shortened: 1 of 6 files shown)


nbdd0121:
This must not be merged until relevant Linux kernel patches land:
* Without https://patchwork.kernel.org/patch/10845957/, Linux will generate SBI call with ASID & size being 0
* Without https://patchwork.kernel.org/patch/10845959/, Linux may call the SBI with a very large range.

nbdd0121:
Also there might be a concurrency issue? Considering the cases where multiple cores call the SBI simultaneously, then one SBI call may override the data set by another, which is clearly undesired. I assume this is why the ipi_type is originally implemented using bit-masks and atomics (actually I originally plan to patch OpenSBI instead of implementing the IPI-based flush for Linux, but as soon as I see the atomic thing there I realised that deeper thought might be necessary to make it correct).

atishp04:
> This must not be merged until relevant Linux kernel patches land:

Initially, I had the same idea. But OpenSBI implementation can be backward compatible with older kernel version because both  SFENCE_VMA and SFENCE_VMA_ASID sbi calls pass zero in a1 & a2 for start and size. We can use this to distinguish between two cases and handle accordingly.

I will update the PR.

atishp04:
> Also there might be a concurrency issue? Considering the cases where multiple cores call the SBI simultaneously, then one SBI call may override the data set by another, which is clearly undesired. I assume this is why the ipi_type is originally implemented using bit-masks and atomics (actually I originally plan to patch OpenSBI instead of implementing the IPI-based flush for Linux, but as soon as I see the atomic thing there I realised that deeper thought might be necessary to make it correct).

Yes. This is a known issue. As you said correctly, the solution requires deeper thought. We are working on fixing this as well.

avpatel:
We do have spinlock_t so using it we can create SMP-safe circular queue on-top-of an array. Maybe something like below:
`
enum sbi_tlb_work_types {
   SBI_TLB_FLUSH_VMA,
   SBI_TLB_FLUSH_VMA_ASID,
   SBI_TLB_FLUSH_VMA_VMID
};

struct sbi_tlb_work {
   unsigned long start;
   unsigned long size;
   unsigned long id; /* id = asid/vmid based on type field */
   unsigned long type;
};

/* Note: Max queue size need not be too large */
#define SBI_TLB_MAX_QUEUE_SIZE 4

struct sbi_tlb_info {
   spinlock_t queue_lock;
   unsigned long queue_head;
   unsigned long queue_tail;
   struct sbi_tlb_work queue[SBI_TLB_MAX_QUEUE_SIZE];
};
`

atishp04:
> > This must not be merged until relevant Linux kernel patches land:
> 
> Initially, I had the same idea. But OpenSBI implementation can be backward compatible with older kernel version because both SFENCE_VMA and SFENCE_VMA_ASID sbi calls pass zero in a1 & a2 for start and size. We can use this to distinguish between two cases and handle accordingly.
> 
> I will update the PR.

Done.

atishp04:
> We do have spinlock_t so using it we can create SMP-safe circular queue on-top-of an array. Maybe something like below:
> `
> enum sbi_tlb_work_types {
> SBI_TLB_FLUSH_VMA,
> SBI_TLB_FLUSH_VMA_ASID,
> SBI_TLB_FLUSH_VMA_VMID
> };
> 
> struct sbi_tlb_work {
> unsigned long start;
> unsigned long size;
> unsigned long id; /* id = asid/vmid based on type field */
> unsigned long type;
> };
> 
> /* Note: Max queue size need not be too large */
> #define SBI_TLB_MAX_QUEUE_SIZE 4
> 
> struct sbi_ipi_data {
> unsigned long ipi_type;
> spinlock_t queue_lock;
> unsigned long queue_head;
> unsigned long queue_tail;
> struct sbi_tlb_work queue[SBI_TLB_MAX_QUEUE_SIZE];
> };
> `
> 
> We can also combine/ignore redundant TLB work at time of inserting in TLB work queue.

Thanks Anup. I will start working on it as a separate patch.

lib/sbi_ipi.c
```diff
@@ -70,6 +80,51 @@ void sbi_ipi_clear_smode(struct sbi_scratch *scratch)
...
+}
+
+static void sbi_ipi_sfence_vma(struct sbi_tlb_info *tinfo)
+{
+	unsigned long start = tinfo->start;
+	unsigned long size = tinfo->size;
+	unsigned long i;
+
+	if (start == 0 && size == 0)
+		sbi_ipi_tlb_flush_all(0);
```
nbdd0121: This is wrong. Flush ASID 0 != global flush. All pages marked with G will now not be flushed at all. And why isn't there a check for SBI_TLB_FLUSH_ALL?

atishp04: Ahh okay. Thanks for pointing out. I misread the spec (got confused x0 with value 0).
I will fix it right now.

avpatel:
> There is an issue with sbi_ipi_sfence_vma. And why is this merged before the concurrency issue is addressed? Even it is going to be submitted as another patch, they should probably go into the master together.

Hi, since your patches are not yet merged and this changes are trying to maintain backward compatibility with older kernels (which as passing a0=0, a1=0, and a2=0). We felt its OK to go ahead with this changes and implement per-HART circular queue as separate patch.

nbdd0121:
> Hi, since your patches are not yet merged and this changes are trying to maintain backward compatibility with older kernels (which as passing a0=0, a1=0, and a2=0). We felt its OK to go ahead with this changes and implement per-HART circular queue as separate patch.

Okay, I see. Currently Linux kernel issues SBI-calls that end up with whole TLB flush anyway, so racing isn't an issue. But I feel it's a bit weird that the correctness of current code relies on the fact that SW doesn't follow the SBI...

avpatel:
> > Hi, since your patches are not yet merged and this changes are trying to maintain backward compatibility with older kernels (which as passing a0=0, a1=0, and a2=0). We felt its OK to go ahead with this changes and implement per-HART circular queue as separate patch.
> 
> Okay, I see. Currently Linux kernel issues SBI-calls that end up with whole TLB flush anyway, so racing isn't an issue. But I feel it's a bit weird that the correctness of current code relies on the fact that SW doesn't follow the SBI...

The releases are our stable points whereas the master branch is our development branch where all PRs and patches are merged. We try our best to keep master branch stable but things can break accidentally.

PR #90, opened 2019-03-12
lib: Allow console message to be quiesced

lbmeng:
By default OpenSBI outputs its banner and some generic details like
platform and firmware on the serial console. Sometimes we don't want
to print such messages hence a new 'QUIESCE' compile time option is
introduced to turn it off.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,7 @@ endif
 GENFLAGS	=	-I$(platform_dir)/include
 GENFLAGS	+=	-I$(platform_common_dir)/include
 GENFLAGS	+=	-I$(include_dir)
+GENFLAGS	+=	$(lib-genflags-y)
 GENFLAGS	+=	$(platform-common-genflags-y)
 GENFLAGS	+=	$(platform-genflags-y)
 GENFLAGS	+=	$(firmware-genflags-y)
--- a/README.md
+++ b/README.md
@@ -68,6 +68,12 @@ directory path, run:
 make O=<build_directory>
 ```
 
+By default OpenSBI outputs its banner and some generic details like platform
+and firmware on the serial console, and this can be quiesced by:
+```
+make QUIESCE=y
+```
+
 To generate files to be installed for using *libsbi.a* in other projects, run:
 ```
 make install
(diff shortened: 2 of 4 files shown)


avpatel:
Downside of having compile time option to suppress OpenSBI prints is that we now have to compile differently so libsbi.a with QUIESCE=1. We should rather have a runtime way of disabling prints.

Further, it's good to suppress the boot-time prints but we should keep the prints coming from errors and warning at runtime.

I think we can have "options" field in "struct sbi_scratch" where firmware linked to libsbi.a can provide options to do things differently. First option can be SBI_SCRATCH_NO_BOOT_PRINTS.

gurjeet:
Just my 2 cents: `quiesce` does not seem to be an appropriate term for the feature being proposed. Something like `no_banner`, `no_startup_messages`, etc. may be more descriptive.

lbmeng:
> Downside of having compile time option to suppress OpenSBI prints is that we now have to compile differently so libsbi.a with QUIESCE=1 is different binary compared to libsbi.a with QUIESCE=0. We should rather have a runtime way of disabling prints.
> 
> Further, it's good to suppress the boot-time prints but we should keep the prints coming from errors and warning at runtime.
> 
> I think we can have "options" field in "struct sbi_scratch" where firmware linked to libsbi.a can provide options to do things differently. First option can be SBI_SCRATCH_NO_BOOT_PRINTS.

Yep, this is a better solution if we add "options" field in "struct sbi_scratch". But how can we change this in run-time? By reading some env variables from ROM?

avpatel:
> > Downside of having compile time option to suppress OpenSBI prints is that we now have to compile differently so libsbi.a with QUIESCE=1 is different binary compared to libsbi.a with QUIESCE=0. We should rather have a runtime way of disabling prints.
> > Further, it's good to suppress the boot-time prints but we should keep the prints coming from errors and warning at runtime.
> > I think we can have "options" field in "struct sbi_scratch" where firmware linked to libsbi.a can provide options to do things differently. First option can be SBI_SCRATCH_NO_BOOT_PRINTS.
> 
> Yep, this is a better solution if we add "options" field in "struct sbi_scratch". But how can we change this in run-time? By reading some env variables from ROM?

Once, we have "options" in "struct sbi_scratch" it will be responsibility of firmware to update it before calling sbi_init().

For OpenSBI reference firmwares, we can set options in firmware/fw_base.S based on optional compile time flags.

Example: You can add optional FW_NO_BOOT_PRINTS compile-flag in firmware/objects.mk. When FW_NO_BOOT_PRINTS is provided to make command-line the firmware/fw_base.S will set SBI_SCRATCH_NO_BOOT_PRINTS in "options" of "struct sbi_scratch"

lbmeng:
See https://github.com/riscv/opensbi/pull/99

Close this one.

PR #96, opened 2019-03-14
platform: plic: Fix comments of programming the interrupt enable register

lbmeng:
The codes that currently program the interrupt enable register for
S-mode disagrees with what the comments say. Fix the comments.

While we are here, add one line comment to describe what is done
for M-mode too.

### DIFF
--- a/platform/common/irqchip/plic.c
+++ b/platform/common/irqchip/plic.c
@@ -86,12 +86,13 @@ int plic_warm_irqchip_init(u32 target_hart,
 	if (plic_hart_count <= target_hart)
 		return -1;
 	
+	/* By default, disable all IRQs for M-mode of target HART */
 	if (m_cntx_id > -1) {
 		for (i = 0; i < ie_words; i++)
 			plic_set_ie(m_cntx_id, i, 0);
 	}
 
-	/* By default, enable all IRQs for S-mode of target HART */
+	/* By default, disable all IRQs for S-mode of target HART */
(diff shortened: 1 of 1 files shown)


platform/common/irqchip/plic.c
```diff
@@ -91,7 +92,7 @@ int plic_warm_irqchip_init(u32 target_hart,
 	/* By default, enable all IRQs for S-mode of target HART */
 	if (s_cntx_id > -1) {
 		for (i = 0; i < ie_words; i++)
-			plic_set_ie(s_cntx_id, i, 0);
+			plic_set_ie(s_cntx_id, i, 0xffffffff);
```
avpatel: We keep the IRQs disabled for S-mode of target HART because S-mode SW (Linux or other OSes) should explicitly enable IRQs when required.

Instead, please fix the comment because it's a typo in comment.

lbmeng: BBL enables all IRQs for S-mode. This is to keep consistency with BBL. A good S-mode SW should disable IRQs during the PLIC initialization, and enable whatever is requested by device drivers. That's how Linux does.

Having said that, I don't have a strong preference for either way.

avpatel: I agree that good S-mode SW should disable IRQs during PLIC initialization but it's not ideal world so keeping IRQs disabled by default is a conservative approach.

PR #98, opened 2019-03-20
platform: Fix PLIC priority base offset

alistair23:
According to the FU540 and E31 manuals the PLIC source priority
address starts at an offset of 0x04 and not 0x00. To aviod confusion
update the address and source offset to match the documentation. This
causes no difference in functionality.

### DIFF
--- a/platform/common/irqchip/plic.c
+++ b/platform/common/irqchip/plic.c
@@ -14,7 +14,7 @@
 #include <plat/tinyfdt.h>
 #include <plat/irqchip/plic.h>
 
-#define PLIC_PRIORITY_BASE		0x0
+#define PLIC_PRIORITY_BASE		0x04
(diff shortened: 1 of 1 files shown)


avpatel:
I had comment similar to Christoph's comment on your Linux patch. I am fine with this patch or just having explicit comments but let's keep priority register address calculation same as Linux driver.

Let's wait for your Linux patch to conclude.

alistair23:
Hopefully we can get this in to Linux then, as this incorrect offset has already caused a bug in OpenSBI and QEMU. I think the current way we handle it is very confusing.

avpatel:
Latest QEMU is now fixed. Do we still need this change?

alistair23:
Yes, this change is improving the readability of the OpenSBI code

alistair23:
Closing as I don't think this is going to be accepted

PR #102, opened 2019-04-01
Sfence race queue

### LINKED ISSUE #87: Improve SBI_REMOTE_SFENCE_VMA_ASID and SBI_REMOTE_SFENCE_VMA implementation
Instead of flushing all TLBs for SBI_REMOTE_SFENCE_VMA_ASID and SBI_REMOTE_SFENCE_VMA we should
flush/sfence only required TLBs.

atishp04:
Fixes #87

### DIFF
--- a/include/sbi/sbi_fifo.h
+++ b/include/sbi/sbi_fifo.h
@@ -1,0 +1,31 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 Western Digital Corporation or its affiliates.
+ *
+ * Authors:
+ *   Atish Patra<atish.patra@wdc.com>
+ *
+ */
+
+#ifndef __SBI_FIFO_H__
+#define __SBI_FIFO_H__
+
+#include <sbi/riscv_locks.h>
+#include <sbi/sbi_types.h>
+#include <sbi/sbi_ipi.h>
+
+struct sbi_fifo {
+	int head;
+	int tail;
+	unsigned long maxsize;
+	spinlock_t qlock;
+	struct sbi_tlb_info queue[SBI_TLB_INFO_MAX_QUEUE_SIZE];
+};
+
+struct sbi_tlb_info * sbi_fifo_dequeue(struct sbi_fifo *fifo);
+int sbi_fifo_enqueue(struct sbi_fifo *fifo, void *data);
+void sbi_fifo_init(struct sbi_fifo *fifo, unsigned long msize);
+bool sbi_fifo_is_empty(struct sbi_fifo *fifo);
+bool sbi_fifo_is_full(struct sbi_fifo *fifo);
+#endif
--- a/include/sbi/sbi_ipi.h
+++ b/include/sbi/sbi_ipi.h
@@ -18,6 +18,8 @@
 #define SBI_IPI_EVENT_SFENCE_VMA_ASID		0x8
 #define SBI_IPI_EVENT_HALT			0x10
 
+#define SBI_TLB_INFO_MAX_QUEUE_SIZE 4
+
 struct sbi_scratch;
 
 struct sbi_ipi_data {
--- a/include/sbi/sbi_ipi.h
+++ b/include/sbi/sbi_ipi.h
@@ -38,6 +38,14 @@ struct sbi_tlb_info {
 	unsigned long type;
 };
 
+inline void sbi_tlb_info_copy(struct sbi_tlb_info *dest, struct sbi_tlb_info *src)
+{
+	dest->start = src->start;
+	dest->size = src->size;
+	dest->asid = src->asid;
+	dest->type = src->type;
+}
+
 int sbi_ipi_send_many(struct sbi_scratch *scratch,
 		      ulong *pmask, u32 event, void *data);
 
--- a/include/sbi/sbi_ipi.h
+++ b/include/sbi/sbi_ipi.h
@@ -19,6 +19,11 @@
 #define SBI_IPI_EVENT_HALT			0x10
 
 #define SBI_TLB_INFO_MAX_QUEUE_SIZE 4
+enum sbi_tlb_info_types {
+	SBI_TLB_FLUSH_VMA,
+	SBI_TLB_FLUSH_VMA_ASID,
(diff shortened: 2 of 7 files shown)


include/sbi/sbi_ipi.h
```diff
@@ -38,6 +38,14 @@ struct sbi_tlb_info {
 	unsigned long type;
 };
 
+inline void sbi_tlb_info_copy(struct sbi_tlb_info *dest, struct sbi_tlb_info *src)
+{
+	dest->start = src->start;
+	dest->size = src->size;
+	dest->asid = src->asid;
+	dest->type = src->type;
+}
```
avpatel: This should not be required for generic sbi_fifo implementation.

atishp04:
> Overall, I think let's make sbi_fifo more generic because it won't impact scratch space usage much.

Yup. That would make a lot of sense. First version of my patch had a generic approach, but I abandoned it because extra pointer wastage in scratch. As you agree that generic one is better, I will bring it back.

include/sbi/sbi_fifo.h
```diff
@@ -0,0 +1,31 @@
...
+	int head;
+	int tail;
+	unsigned long maxsize;
+	spinlock_t qlock;
+	struct sbi_tlb_info queue[SBI_TLB_INFO_MAX_QUEUE_SIZE];
+};
+
+struct sbi_tlb_info * sbi_fifo_dequeue(struct sbi_fifo *fifo);
+int sbi_fifo_enqueue(struct sbi_fifo *fifo, void *data);
+void sbi_fifo_init(struct sbi_fifo *fifo, unsigned long msize);
```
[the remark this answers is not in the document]
atishp04: Let's pass num_entries and entrysize. In that way, the user of fifo doesn't have to calculate the msize.

lib/sbi_fifo.c
```diff
@@ -0,0 +1,84 @@
...
+#include <sbi/sbi_fifo.h>
+
+void sbi_fifo_init(struct sbi_fifo *fifo, unsigned long msize)
+{
+	int i;
+	struct sbi_tlb_info tlb_info_init = {0};
+
+	fifo->head = -1;
+	fifo->tail = -1;
+	fifo->maxsize = msize;
```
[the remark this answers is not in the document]
atishp04: If we pass num_entries in stead of msize, it would be 
fifo->msize = num_entries * entrysize.

include/sbi/sbi_scratch.h
```diff
@@ -37,7 +37,8 @@
 /** Offset of ipi_type in sbi_ipi_data */
 #define SBI_IPI_DATA_IPI_TYPE_OFFSET		(15 * __SIZEOF_POINTER__)
 
-#define SBI_SCRATCH_TLB_INFO_OFFSET		(16 * __SIZEOF_POINTER__)
+#define SBI_SCRATCH_TLB_INFOQ_OFFSET		(16 * __SIZEOF_POINTER__)
+
```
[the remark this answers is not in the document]
atishp04: This would create 32 bytes of gap in scratch. How about just keeping queue right after where the header ends.

#define SBI_SCRATCH_FIFO_HEADER_SIZE sizeof(struct sbi_fifo)
#define SBI_SCRATCH_TLB_QUEUE_MEM_OFFSET (SBI_SCRATCH_TLB_QUEUE_HEAD_OFFSET + SBI_SCRATCH_FIFO_HEADER_SIZE)

We don't have to assign queue memory by hand as well.

atishp04:
I agree with other suggestions as well.

lib/sbi_ipi.c
```diff
@@ -36,10 +36,21 @@ static int sbi_ipi_send(struct sbi_scratch *scratch, u32 hartid,
...
+			 * There may be case where target hart is also
+			 * enqueue in source hart's fifo. Both hart may busy
+			 * loop leading to a deadlock.
+			 * TODO: Introduce a wait/wakeup event mechansim to handle
+			 * this properly.
+			 */
+			__asm__ __volatile("nop");
+			__asm__ __volatile("nop");
+			ret = sbi_fifo_enqueue(ipi_tlb_fifo, data);
+		}
```
gurjeet: Instead of duplicating the call to `sbi_fifo_enqueue()`, I think the while loop can be better written as:

```
while (sbi_fifo_enqueue(ipi_tlb_fifo, data) < 0) {
    __asm__ code
}
```

lib/sbi_ipi.c
```diff
@@ -151,10 +164,15 @@ void sbi_ipi_process(struct sbi_scratch *scratch)
...
 		case SBI_IPI_EVENT_SFENCE_VMA_ASID:
-			sbi_ipi_sfence_vma_asid(sbi_tlb_info_ptr(scratch));
+			tinfo = sbi_fifo_dequeue(ipi_tlb_fifo);
+			while (tinfo) {
+				if (tinfo->type == SBI_TLB_FLUSH_VMA)
+					sbi_ipi_sfence_vma(tinfo);
+				else if (tinfo->type == SBI_TLB_FLUSH_VMA_ASID)
+					sbi_ipi_sfence_vma_asid(tinfo);
+				tinfo = sbi_fifo_dequeue(ipi_tlb_fifo);
+			}
```
gurjeet: Similar to above, instead of duplicating the call to `sbi_fifo_dequeue()`, I think the while loop can be better written as:

```
while (sbi_fifo_dequeue(ipi_tlb_fifo)) {
    if condition
}
```

lib/sbi_ipi.c
atishp04: Done.

atishp04:
Addressed all the comments and raised another PR #105.
Closing this one.

gurjeet:
@atishp04 Just in case you didn't know... You didn't really have to open a new PR to address the review comments. If you push additional commits to the branch that this PR was created from, GitHub will automatically treat your PR as updated by adding "atishp04 pushed new commits to this PR".

atishp04:
@gurjeet I usually do that. However, I don't prefer that if the PR is significantly changed which was true in this case. If you keep on updating same PR with a lot of changes, it is difficult to track the previous changes which may be important in my opinion.

PR #104, opened 2019-04-02
firmware: Fix source fdt alignment

wxjstz:
When I tried to start opensbi with coreboot, I found that aligning to a 16-byte
boundary would make a copy error.

The current riscv platform does not have a 128-bit machine, and there is no need
to align the 16-byte boundary. Corrected to align to an 8-byte boundary.

### DIFF
--- a/firmware/fw_base.S
+++ b/firmware/fw_base.S
@@ -119,7 +119,7 @@ _prev_arg1_override_done:
 	 */
 	beqz	a1, _fdt_reloc_done
 	/* Mask values in a3 and a4 */
-	li	a3, ~0xf
+	li	a3, ~(__SIZEOF_POINTER__ - 1)
 	li	a4, 0xff
 	/* t1 = destination FDT start address */
 	add	s0, a0, zero


firmware/fw_base.S
```diff
@@ -119,7 +119,7 @@ _prev_arg1_override_done:
 	 */
 	beqz	a1, _fdt_reloc_done
 	/* Mask values in a3 and a4 */
-	li	a3, ~0xf
+	li	a3, ~0x7
```
avpatel: Thanks for catching.

How about (__SIZEOF_POINTER__ - 1) instead of ~0x7 ??

wxjstz: Thanks, this suggestion is great.

firmware/fw_base.S
```diff
@@ -119,7 +119,7 @@ _prev_arg1_override_done:
 	 */
 	beqz	a1, _fdt_reloc_done
 	/* Mask values in a3 and a4 */
-	li	a3, ~0xf
+	li	a3, ~(__SIZEOF_POINTER - 1)
```
avpatel: Typo here. Use __SIZEOF_POINTER__ instead of __SIZEOF_POINTER

wxjstz: Done

PR #105, opened 2019-04-03
Sfence race fifo v2

### LINKED ISSUE #87: Improve SBI_REMOTE_SFENCE_VMA_ASID and SBI_REMOTE_SFENCE_VMA implementation

atishp04:
Fixes #87. 
I will move platform/libc/ to lib/sbi_libc as a separate patch so that platform and libsbi can be built independently.

### DIFF
--- a/include/sbi/sbi_fifo.h
+++ b/include/sbi/sbi_fifo.h
@@ -1,0 +1,32 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 Western Digital Corporation or its affiliates.
+ *
+ * Authors:
+ *   Atish Patra<atish.patra@wdc.com>
+ *
+ */
+
+#ifndef __SBI_FIFO_H__
+#define __SBI_FIFO_H__
+
+#include <sbi/riscv_locks.h>
+#include <sbi/sbi_types.h>
+
+struct sbi_fifo {
+	int head;
+	int tail;
+	spinlock_t qlock;
+	unsigned long entrysize;
+	unsigned long num_entries;
+	void *queue;
+};
+
+int sbi_fifo_dequeue(struct sbi_fifo *fifo, void *data);
+int sbi_fifo_enqueue(struct sbi_fifo *fifo, void *data);
+void sbi_fifo_init(struct sbi_fifo *fifo, unsigned long entries,
+		   unsigned long entrysize);
+bool sbi_fifo_is_empty(struct sbi_fifo *fifo);
+bool sbi_fifo_is_full(struct sbi_fifo *fifo);
+#endif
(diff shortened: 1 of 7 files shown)


gurjeet:
To create a link between PRs... the previous version of this PR is at #102.

gurjeet:
This PR addresses the concerns I had in #102, so LGTM.

nbdd0121:
For the FIFO, why isn't it implemented as a ring buffer? The memcpy looks unnecessary.

As for the case of FIFO full, probably it is better to clear the FIFO and enqueue a full flush instead (this would require the lock to be moved out from FIFO).

avpatel:
It is a ring buffer.

The memcpy is because it is a generic FIFO dealing with objects of arbitrary size.

Also, good suggestion on FIFO full case. Atish has other optimizations as well. He is working on the optimizations so expect a patch/PR soon.

nbdd0121:
@avpatel You're right. I only did a quick glance and missed the second part of full check.

nbdd0121:
Just now I had a case where it hit the two "nop" part of code, and deadlocked. So the probability probably isn't low enough to ignore safely.

atishp04:
> Just now I had a case where it hit the two "nop" part of code, and deadlocked. So the probability probably isn't low enough to ignore safely.

Hmm. That happened sooner than I expected. Can you please give the detailed steps to reproduce ?

nbdd0121:
I ran into this randomly when testing #111 using a set of benchmarks. I found out it deadlocks there by opening QEMU's console and do `info registers` on all CPUs, finding out the PCs, and looking into the disassembly of OpenSBI.

I haven't managed to reproduce it with up-to-date OpenSBI, but without #111 is much easier to reproduce. The issue can be workaround by adding an sbi_fifo_inplace_update that replaces one of the entry with a full flush and return.

BTW I found that the #111's use of SBI_FIFO_RESET is incorrect because it does not atomically enqueue the full flush entry while clearing the rest. The caller of SBI expects the entries to be actually flushed by the time it returns, and not atomically enqueue the full flush may invalidate the guarantee.

atishp04:
> adding



> I ran into this randomly when testing #111 using a set of benchmarks. I found out it deadlocks there by opening QEMU's console and do `info registers` on all CPUs, finding out the PCs, and looking into the disassembly of OpenSBI.
> 
Can you provide some details about the benchmarks ? I would like to try them.
I saw fifo full condition only few times in my benchmark even though it was generating more than 1M remote tlb flushes.

> I haven't managed to reproduce it with up-to-date OpenSBI, but without #111 is much easier to reproduce. The issue can be workaround by adding an sbi_fifo_inplace_update that replaces one of the entry with a full flush and return.
> 
That will work but that may be too aggressive. If a genuine workload filling up the fifo and creating a deadlock, we should have a permanent solution.

> BTW I found that the #111's use of SBI_FIFO_RESET is incorrect because it does not atomically enqueue the full flush entry while clearing the rest. The caller of SBI expects the entries to be actually flushed by the time it returns, and not atomically enqueue the full flush may invalidate the guarantee.

Yeah. Thanks for pointing that out. I will fix it.

PR #111, opened 2019-04-03
Tlb flush optimizations

atishp04:


atishp04:
Some performance numbers comparison to IPI mechanism:

Benchmark used: A microbenchmark that mmap a ramdisk (1G) and multiple threads access 50MB of memory randomly.

https://github.com/westerndigitalcorporation/hmmap/blob/master/userspace/hmmap_uspace_common.c

The result is averaged over 25 iterations for 8 threads on HiFive Unleashed board. In both cases around ~1M remote tlb flushes are triggered.

                                                IPI              |    SBI        |          Gain
----------------------------------------------------------------------
Average Write Time             2.53183 .      2.43263           +4.34%
Average Read Time             1.32198         1.24643 .         +6.09%
Total Time                            97.7589         92.859            +5.01%

PR #113, opened 2019-04-09
Allow building platform out-of-tree

avpatel:


avpatel:
Already reviewed on OpenSBI mailing list.

PR #114, opened 2019-04-16
lib: Redirect unhandled traps from non-M modes to S mode

mickflemm:
In case we didn't handle a trap with one of the available
handlers, check if the trap comes from S or U mode and
redirect it to S mode's trap handler.

Signed-off-by: Nick Kossifidis <mick@ics.forth.gr>

### DIFF
--- a/lib/sbi_trap.c
+++ b/lib/sbi_trap.c
@@ -142,6 +142,7 @@ void sbi_trap_handler(struct sbi_trap_regs *regs,
 	const char *msg = "trap handler failed";
 	u32 hartid = sbi_current_hartid();
 	ulong mcause = csr_read(CSR_MCAUSE);
+	ulong mtval = csr_read(CSR_MTVAL);
 
 	if (mcause & (1UL << (__riscv_xlen - 1))) {
 		mcause &= ~(1UL << (__riscv_xlen - 1));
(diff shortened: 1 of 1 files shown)


lib/sbi_trap.c
```diff
@@ -178,6 +179,8 @@ void sbi_trap_handler(struct sbi_trap_regs *regs,
 		msg = "ecall handler failed";
 		break;
 	default:
+		/* If the trap came from S or U mode, redirect it there */
+		rc = sbi_trap_redirect(regs, scratch, regs->mepc, mcause, mtval);
```
avpatel: This sbi_trap_redirect() call should under

```
if (((regs->mstatus & MSTATUS_MPP) >> MSTATUS_MPP_SHIFT) != PRV_M) {
...
}

```

avpatel:
BTW, sorry for delayed response.

We have updated docs/contributing.md to promote patches send via OpenSBI mailing list. In future, please prefer sending to OpenSBI mailing list.

lib/sbi_trap.c
mickflemm: From sbi_trap_redirect:

/* Sanity check on previous mode */
	prev_mode = (regs->mstatus & MSTATUS_MPP) >> MSTATUS_MPP_SHIFT;
	if (prev_mode != PRV_S && prev_mode != PRV_U)
return SBI_ENOTSUPP;

avpatel: Ahh yes, I forgot. We are already checking this in sbi_trap_redirect.

PR #119, opened 2019-04-25
platform/common/libfdt: fix unnecessary multiple inclusions

wxjstz:


avpatel:
Strange....

Why do we need to include libfdt_env.h ?

Do you have a case where you see compile error because of this?

wxjstz:
/home/merle/workspaces/opensbi/platform/common/libfdt/fdt.h:58:2: error: unknown type name 'fdt32_t'

PR #121, opened 2019-04-25
feature: update scratch.next_addr by fdt

wxjstz:
The opensbi, kernel and initramfs may be dynamically loaded by the bootloader,
at which point the bootloader needs to pass the kernel's starting address to
opensbi. Can be passed through the "opensbi,next-addr" of the fdt's "chosen"
node.


avpatel:
First of all, we don't want any FDT/ACPI related code in generic library. We have decided to keep all FDT/ACPI related stuff to be platform specific in OpenSBI. Due to this it is platform support code that explicitly enables LIBFDT if it wants to use it. This way we keep generic library independent of FDT/ACPI standards. Based on this both include/sbi/sbi_fdt.h and lib/sbi_fdt.c are not acceptable hence NACK to this approach.

Secondly, we have deliberately kept all libfdt related headers in platform/common/libfdt directory itself so that later it's easier for us to upgrade libfdt sources from external libfdt. Hence even these changes are not acceptable.

Now, we definitely want to have a mechanism pass next_addr from previous booting stage but we want to do it in FDT/ACPI independent way.

Currently, we have two types of firmware:
1. FW_PAYLOAD, where next_addr points to payload of firmware
2. FW_JUMP, where next_addr is a fixed platform specific address

I suggest we create another firmware called FW_DYNAMIC. Previous booting stage will pass next_addr in a2 register to FW_DYNAMIC while a0 will be HARTID and a1 will be FDT address. 

If previous booting stage loads INITRD as well then it will be previous booting stage's responsibility to update FDT /chosen node with location of INITRD.

Adding FW_DYNAMIC should be very straight forward. If you wish I can add it for you.

wxjstz:
**FW_DYNAMIC** is nice !
Looking forward to your submission !!!

avpatel:
I will be sending FW_DYNAMIC patches to OpenSBI mailing list next week.

To join OpenSBI mailing list, visit:

avpatel:
Did FW_DYNAMIC work for you guys?

Please let me know if you have comments ...

avpatel:
I am closing this PR now. The FW_DYNAMIC is already merged.

PR #134, opened 2019-08-06
Update UART base addresses of qemu/sifive_u platform

zongbox:
The UART addresses should be matched the u-series hardware. After QEMU correct these addresses, it must be modified correspondingly in opensbi platform.
There is the patch of QEMU modification on the fly (https://patchew.org/QEMU/1565020823-24223-1-git-send-email-bmeng.cn@gmail.com/1565020823-24223-10-git-send-email-bmeng.cn@gmail.com/)


alistair23:
Don't merge this until the QEMU changes go in.

Once your patchset is merged do we still need this custom QEMU machine?

zongbox:
Yes, agree, we need to wait QEMU's patches are merged. So far, it can't display normally after entering to S-mode because the UART address of fu540 dts in Linux upstream is different with sifive_u machine of QEMU and qemu/sifive platform of openSBI.
Do the custom QEMU machine you mention means sifive_u machine? If so, we still need it because there are different stuff with virt machine, such as HART configuration and UART driver, etc.

lbmeng:
This patch should NOT be merged.

With the QEMU patch series go in, the OpenSBI support of qemu/sifive_u should be dropped. That was my original plan.

lbmeng:
> Yes, agree, we need to wait QEMU's patches are merged. So far, it can't display normally after entering to S-mode because the UART address of fu540 dts in Linux upstream is different with sifive_u machine of QEMU and qemu/sifive platform of openSBI.
> Do the custom QEMU machine you mention means sifive_u machine? If so, we still need it because there are different stuff with virt machine, such as HART configuration and UART driver, etc. For openSBI part, I think that we could only retain the sifive/fu540 platform, this platform should be worked normally on QEMU sifive_u machine.

No we don't need this special qemu/sifive_u platform at all. If you read my QEMU series cover letter, with that patch series the QEMU 'sifive_u' will be able to boot the real HiFive Unleashed firmware plus Linux kernel image out of the box.

zongbox:
Yes. At the tail of my previous comment, I means that we need sifive_u machine in QEMU, but we doesn't need qemu/sifive_u platform in OpenSBI, because sifive/fu540 platform should be worked on sifive_u machine of QEMU. If we are on the same page, I think this PR could be closed.

lbmeng:
> Yes. At the tail of my previous comment, I means that we need sifive_u machine in QEMU, but we doesn't need qemu/sifive_u platform in OpenSBI, because sifive/fu540 platform should be worked on sifive_u machine of QEMU. If we are on the same page, I think this PR could be closed.

Yep, I already sent a patch to OpenSBI ML to drop the PLATFORM qemu/sifive_u support due to QEMU's improvement on emulation fidelity of its 'sifive_u' machine.

http://lists.infradead.org/pipermail/opensbi/2019-August/000335.html

PR #135, opened 2019-08-19
platform: Add Andes AE350 initial support

nylon7:
This commit provides basic support for the AE350 platform.

Signed-off-by: Zong Li <zongbox@gmail.com>
Signed-off-by: Nylon Chen <nylon7@andestech.com>


platform/andes/ae350/plmt.c
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 Andes Technology Corporation
+ *
+ * Authors:
+ *   Zong Li <zong@andestech.com>
+ *   Nylon Chen <zong@andestech.com>
+ */
+
+#include <sbi/riscv_io.h>
```
lbmeng: It looks that the programming model of this PLMT timer is compatible with CLINT, with the difference of timer register offsets. Can we refactor the existing OpenSBI clint driver a little bit, to easily support PLMT?

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,172 @@
...
+
+	plicsw_dev[source_hart].source_id =
+		*(plicsw_dev[source_hart].plicsw_claim);
+}
+
+static inline void plicsw_complete(void)
+{
+	u32 source_hart = sbi_current_hartid();
+
+	*(plicsw_dev[source_hart].plicsw_claim) =
```
lbmeng: use writel()?

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,172 @@
...
+
+static inline u32 plicsw_get_pending(u32 source_hart, u32 target_hart)
+{
+	return *(plicsw_dev[source_hart].plicsw_pending)
+	       & (PLICSW_HART_MASK >> target_hart);
+}
+
+static inline void plic_sw_pending(u32 target_hart)
+{
+	/* The pending array registers are w1s type.
```
lbmeng: nits: wrong multi-line comment format

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,172 @@
...
+			enable[0] = enable_mask;
+			enable_mask >>= 1;
+		}
+	}
+
+	/* Figure-out PLICSW IPI register address */
+	plicsw_ipi_hart_count = hart_count;
+
+	for (u32 hartid = 0; hartid < AE350_HART_COUNT; hartid++) {
+		if (AE350_ENABLED_HART_MASK & (1<<hartid)) {
```
lbmeng: nits: need space around <<

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,172 @@
...
+
+	/* Clear PLICSW IPI */
+	plicsw_ipi_clear(hartid);
+
+	return 0;
+}
+
+int plicsw_cold_ipi_init(unsigned long base, u32 hart_count)
+{
+
```
lbmeng: nits: remove this blank line

platform/andes/ae350/objects.mk
```diff
@@ -0,0 +1,14 @@
...
+#
+# Authors:
+#   Zong Li <zong@andestech.com>
+#
+
+platform-objs-y += trigger.o plicsw.o plmt.o platform.o
+
+ifdef AE350_ENABLED_HART_MASK
+platform-genflags-y += -DAE350_ENABLED_HART_MASK=$(AE350_ENABLED_HART_MASK)
+endif
```
avpatel: Do you really need AE350_ENABLED_HART_MASK ? If not required then please remove it.

We had added FU540_ENABLED_HART_MASK compile-time option for SiFive FU540 because at that time U-Boot was not SMP capable. After U-Boot SMP support was merged, we have stopped using FU540_ENABLED_HART_MASK on SiFive FU540 as well.

If you have M-mode only HARTs (i.e. HARTs without MMU) which cannot run Linux then you can force disable such HARTs directly in disabled_hart_mask without any compile-time AE350_ENABLED_HART_MASK option.

platform/andes/ae350/platform.c
```diff
@@ -0,0 +1,213 @@
...
+
+	.console_init = ae350_console_init,
+	.console_putc = uart8250_putc,
+	.console_getc = uart8250_getc,
+
+	.irqchip_init = ae350_irqchip_init,
+
+	.ipi_init     = ae350_ipi_init,
+	.ipi_send     = plicsw_ipi_send,
+	.ipi_sync     = plicsw_ipi_sync,
```
avpatel: "ipi_sync" is no longer available in "struct sbi_platform_operations"

We recently removed ipi_sync for all platforms because the generic code now uses atomics to synchronize for remote TLB flushes.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,172 @@
...
+void plicsw_ipi_send(u32 target_hart)
+{
+	if (plicsw_ipi_hart_count <= target_hart)
+		return;
+
+	/* Set PLICSW IPI */
+	plic_sw_pending(target_hart);
+}
+
+void plicsw_ipi_sync(u32 target_hart)
```
avpatel: Please remove plicsw_ipi_sync(). It's not required in latest OpenSBI.

platform/andes/ae350/plmt.c
```diff
@@ -0,0 +1,95 @@
...
+{
+#if __riscv_xlen == 64
+	return readq_relaxed(plmt_time_val);
+#else
+	u64 tmp;
+
+	tmp = readl_relaxed((void *)plmt_time_val + 0x04);
+	tmp <<= 32;
+	tmp |= readl_relaxed(plmt_time_val);
+	return tmp;
```
avpatel: Reading 64bit free running MMIO timer can be tricky on 32bit system because you have to consider higher 32bits can change by the time we read lower 32bits.

This can be addressed using "do { } while" as follows:

```
u32 lo, hi;
do {
    hi = readl_relaxed((void *)plmt_time_val + 0x04);
    lo = readl_relaxed(plmt_time_val);
} while (hi != readl_relaxed((void *)plmt_time_val + 0x04));

return( (u64)hi << 32) | (u64)lo;

```

avpatel: This needs to be fixed in clint_timer_value() as well.

platform/andes/ae350/plmt.c
avpatel: I agree with Bin's comment. PLMT is indeed very similar to CLINT.

We can extend clint_cold_timer_init() to make it suitable for AE350 SOC family. The clint_cold_timer_init() will required addition parameter "type". Based on this "type" parameter we can choose appropriate register offsets. The "type" can be 0 = SiFive OR 1 = AndesTech.

lbmeng: Instead of requiring an additional parameter "type" for clint_cold_timer_init(), we can add two more parameters "time_val_off" and "time_cmp_off" so that the clint timer driver can be reused for Andes PLMT timer.

docs/platform/andes-ae350.md
```diff
@@ -0,0 +1,25 @@
...
+----------------
+
+```
+CROSS_COMPILE=riscv64-linux-  make PLATFORM=andes/ae350 AE350_ENABLED_HART_MASK=0x3
+```
+
+Building Andes AE350 Platform
+-----------------------------
+
+**Linux Kernel Payload**
```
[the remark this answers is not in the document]
NonerKao: Our Linux tree is at https://github.com/andestech/linux/tree/ast-v3_2_0-release-public,
and the DTSes are included there.

We will update this paragraph, thanks!

docs/platform/platform.md
```diff
@@ -24,6 +24,9 @@ OpenSBI currently supports the following virtual and hardware platforms:
 * **Ariane FPGA SoC**: Platform support for the Ariane FPGA SoC used on
   Genesys 2 board.
 
+* **Andes AE350 SoC**: Platform support for the Andes FPGA SoC used on
```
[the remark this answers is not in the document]
nylon7: This platform support for the Andes's SoC (AE350)

I shouldn't have  mentioned FPGA

platform/andes/ae350/plmt.c
NonerKao: Although they look similar in this stage, we are not intended to be compatible with CLINT.  We handle SW (in plicsw.c) and timer (PLMT mentioned here) interrupts separately in Andes AE350 platform, while CLINT is responsible for them both.

avpatel: Fair enough, you can keep PLMT driver in your platform support code.

If you are going to continue with PLMT and PLICSW for future platforms as well then you can also consider moving PLMT and PLICSW to lib/utils/sys (just like CLINT).

NonerKao: While that's a good suggestion, I believe we will keep AE350 that way for a long time and there will be no other platforms in any near future.  Once we have our next platforms that share such design, we will re-form the sources for a better structure.

Thanks anyway! :)

platform/andes/ae350/objects.mk
nylon7: We possibly not support u-boot, so we tend to keep.

platform/andes/ae350/trigger.c
```diff
@@ -0,0 +1,287 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 Andes Technology Corporation
+ *
+ * Authors:
+ *   Zong Li <zong@andestech.com>
+ *   Nylon Chen <zong@andestech.com>
+ */
+
```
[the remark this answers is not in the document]
nylon7: I will drop it for now.

@lbmeng  @avpatel 
Will debugging SBI call be supported in the future?

platform/andes/ae350/objects.mk
avpatel: Okay, no problem.

platform/andes/ae350/plmt.c
avpatel: Sure, no problem.

platform/andes/ae350/objects.mk
nylon7: Sorry, we change our mind.

We fellow your advice.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,145 @@
...
+static u32 plicsw_ipi_hart_count;
+static struct plicsw plicsw_dev[AE350_HART_COUNT];
+
+static inline void plicsw_claim(void)
+{
+	u32 source_hart = sbi_current_hartid();
+
+	volatile uint32_t plicw_claim =
+		*(plicsw_dev[source_hart].plicsw_claim);
+	writel(plicw_claim, &plicsw_dev[source_hart].source_id);
```
avpatel: "source_id" is not a register so no need for writel() here.

platform/andes/ae350/plicsw.c
avpatel: Use writel() here when accessing plicsw_claim register.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,145 @@
...
+
+static u32 plicsw_ipi_hart_count;
+static struct plicsw plicsw_dev[AE350_HART_COUNT];
+
+static inline void plicsw_claim(void)
+{
+	u32 source_hart = sbi_current_hartid();
+
+	volatile uint32_t plicw_claim =
+		*(plicsw_dev[source_hart].plicsw_claim);
```
avpatel: Use readl() here when accessing plicsw_claim register.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,145 @@
...
+{
+	u32 source_hart = sbi_current_hartid();
+
+	*(plicsw_dev[source_hart].plicsw_claim) =
+		plicsw_dev[source_hart].source_id;
+}
+
+static inline u32 plicsw_get_pending(u32 source_hart, u32 target_hart)
+{
+	return *(plicsw_dev[source_hart].plicsw_pending)
```
avpatel: Use readl() here when accessing plicsw_pending register.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,145 @@
...
+	 * -----------------------------------------------
+	 * The bit 7 is used to send IPI to hart 0
+	 * The bit 6 is used to send IPI to hart 1
+	 * The bit 5 is used to send IPI to hart 2
+	 * The bit 4 is used to send IPI to hart 3
+	 */
+	u32 source_hart = sbi_current_hartid();
+	u32 target_offset = (PLICSW_PENDING_PER_HART - 1) - target_hart;
+	u32 per_hart_offset = PLICSW_PENDING_PER_HART * source_hart;
+	*(plicsw_dev[source_hart].plicsw_pending) =
```
avpatel: Use writel() here when accessing plicsw_pending register.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,145 @@
...
+	return 0;
+}
+
+int plicsw_cold_ipi_init(unsigned long base, u32 hart_count)
+{
+	/* Setup source priority */
+	uint32_t *priority = (void *)base + PLICSW_PRIORITY_BASE;
+
+	for (int i = 0; i < AE350_HART_COUNT*PLICSW_PENDING_PER_HART; i++)
+		priority[i] = 1;
```
avpatel: Use writel() here when accessing PLICSW priority registers.

platform/andes/ae350/plicsw.c
```diff
@@ -0,0 +1,145 @@
...
+
+	/* Setup target enable.*/
+	uint32_t enable_mask = PLICSW_HART_MASK;
+
+	for (int i = 0; i < AE350_HART_COUNT; i++) {
+		uint32_t *enable = (void *)base + PLICSW_ENABLE_BASE
+			+ PLICSW_ENABLE_PER_HART * i;
+		enable[0] = enable_mask;
```
avpatel: Use writel() here when accessing PLICSW enable register.

platform/andes/ae350/config.mk
```diff
@@ -0,0 +1,36 @@
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Copyright (c) 2019 Western Digital Corporation or its affiliates.
```
avpatel: Please change this Copyright. It should be copyright by AndesTech right?

platform/andes/ae350/objects.mk
```diff
@@ -0,0 +1,11 @@
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Copyright (c) 2019 Western Digital Corporation or its affiliates.
```
avpatel: Please change this Copyright. It should be copyright by AndesTech right?

platform/andes/ae350/platform.c
```diff
@@ -0,0 +1,194 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 Western Digital Corporation or its affiliates.
```
avpatel: Please change this Copyright. It should be copyright by AndesTech right?

platform/andes/ae350/config.mk
nylon7: Yes, thank you for your reminder.

platform/andes/ae350/trigger.c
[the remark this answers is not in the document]
zongbox: In debug spec, the trigger modules can be accessed by m-mode and d-mode, I had pointed out this in previous proposal whether we need to define a SBI ID for it.
So far, there are no relevant SBI interface to use trigger modules in Linux mainline, you may implement it at the same time if needed.
Otherwise, before the debugging SBI calls becomes standard SBI extension, you can use vendor specific SBI calls when opensbi implementation providing SBI 0.2 spec.

PR #153, opened 2020-01-23
Update UART base addresses for qemu/sifve_u

dalegr:
In the QEMU [1] there was a change of the UART base addresses for
sifive_u machine to match the hardware. Make corresponding changes in
the opensbi for qemu/sifive_u platform.

[1] https://git.qemu.org/?p=qemu.git;a=commitdiff;h=4b55bc2b5f7ff065da5d2b813ee5153c598d3764


dalegr:
I am sorry, I just read the OpenSBI Contribution Guideline and it is seems that I need to add "platform:" to the commit subject and Signed-off-by tag. Do I need to resubmit the Pull Request with changes mentioned above?

avpatel:
We generally don't prefer GitHub based PR reviews because lot of the folks interested/watching OpenSBI development are on OpenSBI mailing list. Even if we review your patch here, we still have to send it to OpenSBI mailing list.

We have two options here:
1. You join OpenSBI mailing list and send patch to OpenSBI mailing list. The patch has to be as-per guideline having appropriate subject prefix and signed-off-by
2. We review it here and then I send patch to OpenSBI mailing list on your behalf.

To join OpenSBI mailing list, visit

dalegr:
Hello @avpatel

Thank you for the answer! I will send this patch to the mailing list.

alistair23:
We are planning on removing the qemu/sifive_u platform as QEMU is compatible with the Unleashed platform. You should just use the Unleashed platform from OpenSBI

PR #155, opened 2020-01-31
makefile: add support for building on macOS

arichardson:
On macOS the readlink command does not include a -f flag. Instead default
to using GNU readlink (which is often installed as greadlink).

I am not sure why readlink -f is being used here instead of $(realpath ).
If that is also an acceptable solution I'll update the PR to use that instead.

Signed-off-by: Alex Richardson <Alexander.Richardson@cl.cam.ac.uk>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,18 @@
 # o  Do not print "Entering directory ...";
 MAKEFLAGS += -r --no-print-directory
 
+# readlink -f requires GNU readlink
+# FIXME: could these calls be changed to use $(realpath ...) instead?
+ifeq ($(shell uname -s),Darwin)
+READLINK ?= greadlink
+else
+READLINK ?= readlink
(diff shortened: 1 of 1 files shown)


avpatel:
Overall patch looks good. You can "-r" option to READLINK make variable definition because greadlink does not have it.

Can you send your patches to OpenSBI mailing list using "git send-email"?

To join OpenSBI mailing list, visit:

Regards,
Anup

avpatel:
I have send patch to OpenSBI mailing list on behalf of you.

The use of $(realpath) in Makefile can be separate patch. You can send a patch for this to OpenSBI mailing list.

avpatel:
This patch is now merged.

Please join OpenSBI mailing list and if possible send patches directly to OpenSBI mailing list.

Regards,
Anup

PR #156, opened 2020-01-31
platform: Add Spike initial support

jrtc27:
Signed-off-by: James Clarke <jrtc27@jrtc27.com>


jrtc27:
Forwarded to the mailing list as http://lists.infradead.org/pipermail/opensbi/2020-January/001058.html

PR #161, opened 2020-04-21
platform: Add nuclei ux600 platform

fanghuaqi:
This PR is used to adding support for Nuclei UX600 RISC-V 64bit Core.

> Nuclei UX600 is a riscv 64bit core developed by Nuclei System Technology, see https://nucleisys.com/product.php

### DIFF
--- a/lib/utils/sys/nuclei_timer.c
+++ b/lib/utils/sys/nuclei_timer.c
@@ -1,0 +1,159 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2020 Nuclei Corporation or its affiliates.
+ *
+ * Authors:
+ *   lujun <lujun@nucleisys.com>
+ */
+
+#include <sbi/riscv_asm.h>
+#include <sbi/riscv_atomic.h>
+#include <sbi/riscv_io.h>
+#include <sbi_utils/sys/nuclei_timer.h>
+
+static u32 nuclei_timer_ipi_hart_count;
+static volatile void *nuclei_timer_ipi_base;
+static volatile u32 *nuclei_timer_ipi;
+
+void nuclei_timer_ipi_send(u32 target_hart)
+{
+	if (nuclei_timer_ipi_hart_count <= target_hart)
+		return;
+
+	/* Set RV_TIMER IPI */
+	writel(1, &nuclei_timer_ipi[target_hart]);
+}
+
+void nuclei_timer_ipi_clear(u32 target_hart)
+{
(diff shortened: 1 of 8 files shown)


lib/utils/sys/nuclei_timer.c
```diff
@@ -0,0 +1,159 @@
...
+
+int nuclei_timer_cold_timer_init(unsigned long base, u32 hart_count,
+			  bool has_64bit_mmio)
+{
+	/* Figure-out RV_TIMER Time register address */
+	nuclei_timer_time_hart_count		= hart_count;
+	nuclei_timer_time_base			= (void *)base;
+	nuclei_timer_time_val			= (u64 *)(nuclei_timer_time_base + 0x0);
+	nuclei_timer_time_cmp			= (u64 *)(nuclei_timer_time_base + 0x8);
+
```
atishp04, line 149: It looks like this timer block is identical to a clint block except the offset for timeval and timecmp.
If that's the case, generic clint can be improved in such a way that, ux600 can use clint block as it is.

lib/utils/sys/objects.mk
```diff
@@ -7,5 +7,6 @@
 #   Anup Patel <anup.patel@wdc.com>
 #
 
+libsbiutils-objs-y += sys/nuclei_timer.o
```
atishp04, line 10: As nuclei_timer is very similar to clint, we should try to reuse the code.

platform/nuclei/ux600/platform.c
```diff
@@ -0,0 +1,176 @@
...
+	return plic_warm_irqchip_init(hartid, (hartid) ? (2 * hartid - 1) : 0,
+				      (hartid) ? (2 * hartid) : -1);
+}
+
+static int ux600_ipi_init(bool cold_boot)
+{
+	int rc;
+
+	if (cold_boot) {
+		rc = nuclei_timer_cold_ipi_init(UX600_RV_TIMER_ADDR + 0xffc, UX600_HART_COUNT);
```
atishp04, line 115: please define a macro for 0xffc as well.

lib/utils/sys/nuclei_timer.c
fanghuaqi: Thanks for your review, I will check and see how to easily reuse with clint code.

fanghuaqi:
Hi @atishp04 , I talked with our hw engineer, our nuclei timer will also be compabable with clint  timer, I also update our opensbi portable code, I will close this PR, and use new PR #162 to  track it.

Thanks
Huaqi

PR #164, opened 2020-04-29
firmware: fw_base: Remove CSR_MTVEC update check

### LINKED ISSUE #163: why csrr MTVEC
Hi,

Basicaly i was trying to strip down the CSR requirements of openSBI to make a softcore smaller/faster, and hit the following : 

https://github.com/riscv/opensbi/blob/e5a7f556ce22984128bed0a336551a968c818450/firmware/fw_base.S#L404

Is there some CPU in the wildness which do not update MTVEC properly ?
I'm asking to understand why that test was implemented.

Regards
Charles

Dolu1990:
Fix #163

### DIFF
--- a/firmware/fw_base.S
+++ b/firmware/fw_base.S
@@ -400,9 +400,6 @@ _start_warm:
 	/* Setup trap handler */
 	la	a4, _trap_handler
 	csrw	CSR_MTVEC, a4
-	/* Make sure that mtvec is updated */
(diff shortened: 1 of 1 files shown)


atishp04:
Thanks for sending the patch to the mailing list as well. I have reviewed it in the mailing list.
We were using Github PRs in initial days of OpenSBI but we prefer mailing list based patch reviews as more opensbi community members are on the mailing list.

You can close the PR.

PR #167, opened 2020-06-02
lib: Allow the OS to map the regions that are protected by PMP

AlexGhiti:
This is achieved by removing the 'no-map' property from the
'reserved-memory' node when PMP is present, otherwise we keep it as it
offers a small protection if the OS does not map this region at all.

If PMP is present, telling the OS not to map the reserved regions does not
add much protection since it only avoids access to regions that are already
protected by PMP. But by not allowing the OS to map those regions, it
creates holes in the OS system memory map and prevents the use of
hugepages which would generate, among other benefits, less TLB miss.

Signed-off-by: Alexandre Ghiti <alex@ghiti.fr>

### DIFF
--- a/lib/utils/fdt/fdt_fixup.c
+++ b/lib/utils/fdt/fdt_fixup.c
@@ -70,7 +70,7 @@ void fdt_plic_fixup(void *fdt, const char *compat)
 
 static int fdt_resv_memory_update_node(void *fdt, unsigned long addr,
 				       unsigned long size, int index,
-				       int parent)
+				       int parent, bool no_map)
 {
 	int na = fdt_address_cells(fdt, 0);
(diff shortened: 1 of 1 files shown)


atishp04:
For Qemu & unleashed, the PMP reserved memory ranges(opensbi firmware region) are in the beginning of DRAM which kernel doesn't use anyways. For 64bit, kernel is loaded at 2MB offset and configures memory only after DRAM + 2MB.

Do you have a platform where PMP configures some memory after DRAM + 2MB offset ?

It is preferred to discuss the patches in the OpenSBI mailing list[1] rather than PRs.
Can you please send the patch to the mailing list as well?

[1] http://lists.infradead.org/pipermail/opensbi/

avpatel:
This is already merged hence closing this PR.

PR #178, opened 2020-08-13
lib: sbi: relax scoldboot_lock spinlock

xypron:
OpenSBI hangs in wake_coldboot_harts() indefinitely on the MaixDuino board
when trying to boot U-Boot as FW_PAYLOAD.

Even if reading and writing coldboot_done is not atomic it is safe to check
it without using a spinlock.

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>


xypron:
MaixDuino is a Kendryte K210 board. For building U-Boot I used sipeed_maix_bitm_defconfig with CONFIG_SYS_TEXT_BASE=0x80020000.

xypron:
Patch needs rework. Cf. http://lists.infradead.org/pipermail/opensbi/2020-August/000072.html

PR #200, opened 2021-03-12
include: headers: Replace __ASSEMBLY__ with __ASSEMBLER__

maroueneboubakri:
GCC has already a predefined macro `__ASSEMBLER__` therefore, it can be
used without the need to define a new flag with `-D__ASSEMBLY__`.
This is useful when adding the library to projects having a build
system such one can build without the need to make changes.
THe build system does not use the Makefile in the sources tree.

Signed-off-by: Marouene Boubakri <marouene.boubakri@nxp.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,7 +217,7 @@ CPPFLAGS	+=	$(GENFLAGS)
 CPPFLAGS	+=	$(platform-cppflags-y)
 CPPFLAGS	+=	$(firmware-cppflags-y)
 
-ASFLAGS		=	-g -Wall -nostdlib -D__ASSEMBLY__
+ASFLAGS		=	-g -Wall -nostdlib
 ASFLAGS		+=	-fno-omit-frame-pointer -fno-optimize-sibling-calls
 ASFLAGS		+=	-mno-save-restore -mstrict-align
 ASFLAGS		+=	-mabi=$(PLATFORM_RISCV_ABI) -march=$(PLATFORM_RISCV_ISA)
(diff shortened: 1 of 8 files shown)


avpatel:
Please send your patch to OpenSBI mailing list because we have we have mailing list based patch review.

The link to join OpenSBI mailing list is

Regards,
Anup

maroueneboubakri:
> Please send your patch to OpenSBI mailing list because we have we have mailing list based patch review.
> 
> The link to join OpenSBI mailing list is
> http://lists.infradead.org/mailman/listinfo/opensbi
> 
> Regards,
> Anup

Fine but [Contribution Guideline](https://github.com/riscv/opensbi/blob/master/docs/contributing.md)  mentions that contributions could be sent as GitHub Pull Requests (PRs) also.

maroueneboubakri:
Well, after 5 attempts to send the patch via the mailing-list and I receive back and email telling that it has been blocked by the server (last one telling suspicious header), I cancel this. It turned out then, that the mail found its way and we end-up having duplicates. 

**The mailing-lists way for code contribution are really awful and devs are still using it nowadays !**
Github with all its rich features is only used only to host code and that's sad.

Contributing to this projet this way is really hard .. Thank you

avpatel:
You need subscribe to the mailing list and send patch using "git send-email"

Majority of OpenSBI contributors prefer mailing based reviews so we have stopped taking GitHub PRs. I personally don't see any limitation in email based reviews and lot of large open-source projects (Linux kernel, U-Boot, QEMU, etc) still use it.

Regards,
Anup

PR #203, opened 2021-03-26
lib: utils: Add Xilinx uartlite implementation

chenguokai:
Uartlite is a device commonly used on Xilinx platforms. We add OpenSBI support for uartlite so that OpenSBI boots on SoCs that utilize uartlite for output.


avpatel:
We have mailing list based patch review so it would be great if you can send this patch to OpenSBI mailing list.

Regards,
Anup

avpatel:
Already available in upstream hence closing this PR.

PR #205, opened 2021-04-04
Actually allow fw_boot_hart skip the lottery.

ahmedcharles:


avpatel:
We can't skip HART lottery because for quite a few platforms all HARTs jump to OpenSBI at the same time and initial setup needs to be done by one HART which could be any HART.

ahmedcharles:
In the existing code, if fw_boot_hart returns `-1`, the lottery is used to determine the 'boot' hart. If it returns something else, then all other harts than that number are sent to `_wait_relocate_copy_done` and only the hart matching the right hart id actually participates in the lottery. I originally assumed that resulted in incorrect behavior when initially submitting the PR but I realized that it's just unnecessary work and makes for more confusing code. The only way that the lottery is required when multiple harts jump to OpenSBI at the same time is if multiple harts have the same id, but that's forbidden.

The behavior seems to be correct with or without the change, I just think it's clearer with the change and it does less work.

PR #209, opened 2021-04-29
lib: Fix buffer overflow issue in FDT fixups

zetalog:
We encounter a bug in Linux kernel that some cores are failed to be disabled and some cores are crashed during smpboot. And the root cause of the issue is the SBI FDT fixups.
The FDT fixup functions should modify a device tree node with the value bounded by its original boundary. That's why the official method to make an Linux kernel OF device node unavailable is to make its length to be 0. As strlen("") is surely less than strlen("okay"), and strlen("disabled") breaks the runtime FDT.

This patch fixes the buffer overflow issue in the FDT fixups.

Reference: **linux/drivers/of/base.c __of_device_is_available()**

### DIFF
--- a/include/sbi_utils/fdt/fdt_fixup.h
+++ b/include/sbi_utils/fdt/fdt_fixup.h
@@ -13,7 +13,7 @@
  * Fix up the CPU node in the device tree
  *
  * This routine updates the "status" property of a CPU node in the device tree
- * to "disabled" if that hart is in disabled state in OpenSBI.
+ * to "" if that hart is in disabled state in OpenSBI.
  *
  * It is recommended that platform codes call this helper in their final_init()
  *
--- a/lib/utils/fdt/fdt_domain.c
+++ b/lib/utils/fdt/fdt_domain.c
@@ -142,7 +142,7 @@ static int __fixup_disable_devices(void *fdt, int doff, int roff,
 		if (coff < 0)
 			return coff;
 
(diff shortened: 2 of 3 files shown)


avpatel:
The status = "" is a device DT not node is treated as enabled device.

This change actually breaks the FDT fixups.

Better way to fix your buffer overflow issue is resize FDT during FDT fixups.

zetalog:
No, the actual code is:
{code}
static bool __of_device_is_available(const struct device_node *device)
{
        const char *status;
        int statlen;

        if (!device)
                return false;

        status = __of_get_property(device, "status", &statlen);
        if (status == NULL)
                return true;

        if (statlen > 0) {
                if (!strcmp(status, "okay") || !strcmp(status, "ok"))
                        return true;
        }

        return false;
}
{code}
That means, when statlen is 0, the function returns false (not enabled).
It only returns true when the node doesn't exist or the node exists with string "ok" or "okay".

zetalog:
BTW, the problem I encountered is:
- After using the CPU disabling mechanism, I saw "CPUx failed to come online" errors in the kernel log, where X is not disabled.
- Tracing back, I found some "CPU with hartid=x is not available" errors in the kernel log, where X is not expected.

Unavailable
Online

After applying this fix, all problems are gone.

zetalog:
OK.
Will give it a try and get back with tested fixes.

PR #211, opened 2021-05-08
Makefile: unconditionally disable SSP

ffontaine:
Though `-nostdlib` is passed in `{C,+AS,DTSCPP}FLAGS`, `-fno-stack-protector` must also be passed to avoid linking errors related to undefined references to `__stack_chk_guard` and `__stack_chk_fail` if toolchain enforces -fstack-protector.

Fixes:
 - https://gitlab.com/kubu93/buildroot/-/jobs/1247043359

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,7 +203,7 @@ GENFLAGS	+=	$(libsbiutils-genflags-y)
 GENFLAGS	+=	$(platform-genflags-y)
 GENFLAGS	+=	$(firmware-genflags-y)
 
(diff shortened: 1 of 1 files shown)


avpatel:
We only need "--fstack-no-protector" in CFLAGS. Otherwise looks good to me.

Can you send updated patch to OpenSBI mailing list??

Regards,
Anup

PR #213, opened 2021-05-12
lib: utils: Include sbi_strings.h directly

JohnAZoidberg:
This file uses `sbi_strncpy`, `sbi_strncmp` and `sbi_memset`, which are defined in `sbi_strings.h`.
It should already be transitively included via `#include <sbi/sbi_scratch.h>` but my toolchain doesn't do that for some reason. Perhaps because the functions that refer to them are `static`?

Anyways, it would be good to directly reference the header here.

### DIFF
--- a/lib/utils/fdt/fdt_domain.c
+++ b/lib/utils/fdt/fdt_domain.c
@@ -13,6 +13,7 @@
 #include <sbi/sbi_error.h>
 #include <sbi/sbi_hartmask.h>
 #include <sbi/sbi_scratch.h>
+#include <sbi/sbi_string.h>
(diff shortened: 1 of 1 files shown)


avpatel:
Your patch looks good.

If possible please send it to mailing list

Regards,
Anup

JohnAZoidberg:
Done: http://lists.infradead.org/pipermail/opensbi/2021-May/000982.html

JohnAZoidberg:
This is not the correct solution.
See http://lists.infradead.org/pipermail/opensbi/2021-May/001006.html

PR #215, opened 2021-06-08
doc: clarify make parameter usage

axel-h:
polish the docs a bit and clarify what will work


avpatel:
Can you please re-base and send your patch using "git send-email" to opensbi@lists.infradead.org ?

Regards,
Anup

PR #223, opened 2021-09-20
[RFC] adapting OpenSBI patches for Sifive U740 Unmatched

maquefel:
Hello Atish,

Thank you for your PMU patch series, i've managed to get it work on Sifive U740.

These series is simply proposal for some thoughts:

- it seems that **mcountinhibit** is optional according to current ISA (correct me if i am wrong), 
  also in current state of Linux PMU driver it doesn't give much impact, as registers are written
  on counter start by kernel driver and readied before stopping. Moreover by such a scrict check 
  of **mcountinhibit** in **sbi_pmu** extension, firmware events are also become unavailable.
  
- poorman replacement of hardware mode inhibition for M-Mode only, so we can count cycles, intructions
  spend in M-Mode, and this raises the necessity of introducing "shadow" versions of hardware cycles, 
  intructions counters as real counters require overwriting values on start by Linux PMU driver, so we 
  can't rely on them in our calculations or should provide some sophisticated hooks to update value on write.
  
This is the resulting example of counting cycles, instructions spent in M-Mode:

```
# SBI_PMU_FW_SHADOW_CPU_CYCLES - r8000000000000016
# SBI_PMU_FW_MACHINE_CYCLES - r8000000000000018
perf stat -r 5 --table -a -e r8000000000000016,r8000000000000018 sleep 5

 Performance counter stats for 'system wide' (5 runs):

         186082497      r8000000000000016                                             ( +-  0.19% )
          68059546      r8000000000000018                                             ( +-  0.06% )

         # Table of individual measurements:
         5.1170940 (+0.0000958) #
         5.1170460 (+0.0000478) #
         5.1169850 (-0.0000132) #
         5.1168250 (-0.0001732) #
         5.1170410 (+0.0000428) #

         # Final result:
         5.1169982 +- 0.0000466 seconds time elapsed  ( +-  0.00% )
       
# SBI_PMU_FW_SHADOW_INSTRUCTIONS - r8000000000000017
# SBI_PMU_FW_MACHINE_INSTRUCTIONS - r8000000000000019
perf stat -r 5 --table -a -e r8000000000000017,r8000000000000019 sleep 5

 Performance counter stats for 'system wide' (5 runs):

          16770573      r8000000000000017                                             ( +-  0.75% )
           5270820      r8000000000000019                                             ( +-  0.63% )

         # Table of individual measurements:
         5.1170950 (+0.0000688) #
         5.1170520 (+0.0000258) #
         5.1169750 (-0.0000512) #
         5.1169540 (-0.0000722) #
         5.1170550 (+0.0000288) #

         # Final result:
         5.1170262 +- 0.0000265 seconds time elapsed  ( +-  0.00% )
```

Looking forward to hear opionions on this issue.


atishp04:
@maquefel I thought unmatched already supports mcountinhibit. Isn't it ?

Moreover, what you are suggesting is what legacy PMU driver will do. If you need just cycle/instret on a platform without mcouninhibit, the linux driver should automatically fallback to the legacy driver. You can do perf stat for these two counters. Let me know if it doesn't work.

maquefel:
Hello Atish!

Thanks for looking into this.

> @maquefel I thought unmatched already supports mcountinhibit. Isn't it ?

No, Unmatched doesn't have mcountinhibit. And Sifive states that it's not mandatory.

> Moreover, what you are suggesting is what legacy PMU driver will do. If you need just cycle/instret on a platform without mcouninhibit, the linux driver should automatically fallback to the legacy driver. You can do perf stat for these two counters. Let me know if it doesn't work.

Well i don't need **just** cycle/instret, i need all that can i get of Unmatched PMU without mcouninhibit.

Without OpenSBI PMU extension:
- no firmware counters
- no hardware counters except cycle/insret

AFAIK the only Linux capable RISC-V boards we gave currently are Nezha and Unmatched.

So is **mcouninhibit** really that mandatory ? What was the reason to rely so hard on it?

maquefel:
v1 -> v2:
Fixed issue with "error: label at end of compound statement"
break; after each default.

atishp04:
> was



> Hello Atish!
> 
> Thanks for looking into this.
> 
> > @maquefel I thought unmatched already supports mcountinhibit. Isn't it ?
> 
> No, Unmatched doesn't have mcountinhibit. And Sifive states that it's not mandatory.
> 
> > Moreover, what you are suggesting is what legacy PMU driver will do. If you need just cycle/instret on a platform without mcouninhibit, the linux driver should automatically fallback to the legacy driver. You can do perf stat for these two counters. Let me know if it doesn't work.
> 
> Well i don't need **just** cycle/instret, i need all that can i get of Unmatched PMU without mcouninhibit.
> 
> Without OpenSBI PMU extension:
> 
> * no firmware counters
> * no hardware counters except cycle/insret
> 

I agree. We can relax the firmware counters restriction for sure. SBI PMU extension will be enabled even if mcountinhibit is not present. But it will return zero hardware counters. Only firmware counters can be used.

> AFAIK the only Linux capable RISC-V boards we have currently are Nezha and Unmatched.

Nezha board has a completely incompatible PMU that has bunch of custom CSRs and a local interrupt number that is reserved.
AFAIK, the starlight board has mcountnihibit implemented even though it has the same U74.

> 
> So is **mcouninhibit** really that mandatory ? What was the reason to rely so hard on it?

We made mcountinhibit mandatory because we can't really start/stop the counters without it. Thus, we can't implement the core functionality of SBI PMU extension.

maquefel:
Hello Atish.

> We made mcountinhibit mandatory because we can't really start/stop the counters without it.

Well without it counters can't be stopped/disabled at all - that's no doubt.

> Thus, we can't implement the core functionality of SBI PMU extension.

And that statement worries me. If we looks at how Linux kernel PMU operates counters we will see the following from OpenSBI point of view:

For firmware counters:
- pmu_ctr_start_fw and write value provided by kernel PMU
- pmu_ctr_read_fw
- pmu_ctr_stop_fw

For hardware counter's everything is same except instead of pmu_ctr_read_fw we have a direct read, which can be replaced with pmu_ctr_read_hw to see the whole picture from the same point of view.

Counters are started/stopped in start_hw and stop_hw respectively, so that's not so drastic from my point of view.

Am i missing something ? May be you have a public paper/publications accompanying this implementation ?

atishp04:
> Hello Atish.
> 
> > We made mcountinhibit mandatory because we can't really start/stop the counters without it.
> 
> Well without it counters can't be stopped/disabled at all - that's no doubt.
> 
> > Thus, we can't implement the core functionality of SBI PMU extension.
> 
> And that statement worries me. If we looks at how Linux kernel PMU operates counters we will see the following from OpenSBI point of view:
> 
> For firmware counters:
> 
> * pmu_ctr_start_fw and write value provided by kernel PMU
> * pmu_ctr_read_fw
> * pmu_ctr_stop_fw
> 
> For hardware counter's everything is same except instead of pmu_ctr_read_fw we have a direct read, which can be replaced with pmu_ctr_read_hw to see the whole picture from the same point of view.
> 
> Counters are started/stopped in start_hw and stop_hw respectively, so that's not so drastic from my point of view.

We can improve OpenSBI implementation in such a way that it provides an illusion that counters are actually stopped/started. In reality, it counters won't be stopped in hardware without mcountinhibit.

Let me know think about the cleanest approach to implement this. I will revise the OpenSBI series with that. 

FYI: We don't need additional firmware events once we fix the OpenSBI.

> 
> Am i missing something ? May be you have a public paper/publications accompanying this implementation ?

maquefel:
> We can improve OpenSBI implementation in such a way that it provides an illusion that counters are actually stopped/started. In reality, it counters won't be stopped in hardware without mcountinhibit.

That's totally true. I am not happy of missing **mcountinhibit** either and if it was mandatory by ISA  there would be nothing to talk about.

> Let me know think about the cleanest approach to implement this. I will revise the OpenSBI series with that.
> FYI: We don't need additional firmware events once we fix the OpenSBI.

I suggest moving discussion to mail-lists, not much people hanging around here.

atishp04:
> > We can improve OpenSBI implementation in such a way that it provides an illusion that counters are actually stopped/started. In reality, it counters won't be stopped in hardware without mcountinhibit.
> 
> That's totally true. I am not happy of missing **mcountinhibit** either and if it was mandatory by ISA there would be nothing to talk about.
> 
> > Let me know think about the cleanest approach to implement this. I will revise the OpenSBI series with that.
> > FYI: We don't need additional firmware events once we fix the OpenSBI.
> 
> I suggest moving discussion to mail-lists, not much people hanging around here.

Sounds good to me. I will post the patches to the mailing list once I relax the mcountinhibit requirement. We can discuss further after that.

avpatel:
This is now resolved so closing this PR

PR #227, opened 2021-11-02
SBI: add a workaround for clang

compnerd:
When building with clang with `-Os -ffreestanding`, it seems that clang
improperly forms a call to `memset`.  With `-ffreestanding`, the C
standard as per Clause 4, the compiler cannot necessarily assume that
anything beyond:
- float.h
- iso646.h
- limits.h
- stdalign.h
- stdarg.h
- stdbool.h
- stddef.h
- stdint.h
- stdnoreturn.h
- fenv.h
- math.h
and the numeric conversion functions of stdlib.h.

This change explicitly converts the memset into a `sbi_memset`.  The
overhead here should be minimal, namely an extra jump for the memset
rather than an inline unrolled implementation.

### DIFF
--- a/lib/sbi/sbi_ecall.c
+++ b/lib/sbi/sbi_ecall.c
@@ -98,9 +98,11 @@ int sbi_ecall_handler(struct sbi_trap_regs *regs)
 	struct sbi_ecall_extension *ext;
 	unsigned long extension_id = regs->a7;
 	unsigned long func_id = regs->a6;
-	struct sbi_trap_info trap = {0};
+	struct sbi_trap_info trap;
 	unsigned long out_val = 0;
 	bool is_0_1_spec = 0;
+
+	sbi_memset(&trap, 0, sizeof(trap));
 
 	ext = sbi_ecall_find_extension(extension_id);
 	if (ext && ext->handle) {
--- a/lib/sbi/sbi_hart.c
+++ b/lib/sbi/sbi_hart.c
@@ -337,7 +337,9 @@ done:
 static unsigned long hart_pmp_get_allowed_addr(void)
 {
 	unsigned long val = 0;
-	struct sbi_trap_info trap = {0};
+	struct sbi_trap_info trap;
+
+	sbi_memset(&trap, 0, sizeof(trap));
 
 	csr_write_allowed(CSR_PMPADDR0, (ulong)&trap, PMP_ADDR_MASK);
 	if (!trap.cause) {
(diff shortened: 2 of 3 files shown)


asb:
Is there an upstream bug report for this issue? (I didn't find it from a quick search of LLVM Bugzilla, but I could have missed it!)

compnerd:
No, I've not filed a report, especially with the migration and the state of bugzilla, I figured that it wasn't as valuable.  This doesn't feel like a RISCV specific issue, its a general issue IMO.  I do have some competing priorities, but this is pretty high on my list of things to resolve.  However, even if I resolve it, this is something that prevents the build of OpenSBI with clang with a released compiler, so it seems reasonable to add this workaround for the time being.

asb:
Agreed - I didn't mean to derail this PR review! Thanks for confirming.

compnerd:
@avpatel ping

avpatel:
Sorry for the delay, we are generally more active on mailing list.

Can you please send your patch using "git send-email" to opensbi@lists.infradead.org ?

Regards,
Anup

jrtc27:
Both GCC and Clang rely on memset and memcpy being implemented, the exact heuristics for when to inline each are just different. See https://godbolt.org/z/638G4dbfo.

jrtc27:
I strongly suggest that OpenSBI implement the real memset and memcpy.

jrtc27:
> GCC aims towards being usable as a conforming freestanding implementation, or as the compiler for a conforming hosted implementation. By default, it acts as the compiler for a hosted implementation, defining __STDC_HOSTED__ as 1 and presuming that when the names of ISO C functions are used, they have the semantics defined in the standard. To make it act as a conforming freestanding implementation for a freestanding environment, use the option -ffreestanding; it then defines __STDC_HOSTED__ to 0 and does not make assumptions about the meanings of function names from the standard library, with exceptions noted below. To build an OS kernel, you may well still need to make your own arrangements for linking and startup. See Options Controlling C Dialect.
> 
> GCC does not provide the library facilities required only of hosted implementations, nor yet all the facilities required by C99 of freestanding implementations on all platforms. To use the facilities of a hosted environment, you need to find them elsewhere (for example, in the GNU C library). See Standard Libraries.
> 
> Most of the compiler support routines used by GCC are present in libgcc, but there are a few exceptions. GCC requires the freestanding environment provide memcpy, memmove, memset and memcmp. Finally, if __builtin_trap is used, and the target does not implement the trap pattern, then GCC emits a call to abort.

-- https://gcc.gnu.org/onlinedocs/gcc/Standards.html

avpatel:
> I strongly suggest that OpenSBI implement the real memset and memcpy.

The problem is projects using OpenSBI as library might have their own memcpy/memset which then conflicts with OpenSBI so we have sbi_memcpy/sbi_memset.

Regards,
Anup

jrtc27:
> > I strongly suggest that OpenSBI implement the real memset and memcpy.
> 
> The problem is projects using OpenSBI as library might have their own memcpy/memset which then conflicts with OpenSBI so we have sbi_memcpy/sbi_memset.
> 
> Regards, Anup

So don't put it in libsbi, put it in just the firmware blob. Those projects will be required, due to the same GCC rules, to provide their own memcpy etc anyway.

avpatel:
I will send patch to OpenSBI mailing list myself hence closing this issue.

PR #229, opened 2021-11-19
Ensure reproducible builds regardless of build path

vagrantc:
Upstream commit 12753d22563f7d2d01f2c6644c7b66b06eb5c90f introduced
uses of __FILE__ which may result in the build path getting embedded
into the resulting binary.

The -ffile-prefix-map argument is available in gcc 8 and clang 10, which can be used to strip out the absolute part of the path from the artifacts the compiler produces.

https://reproducible-builds.org/docs/build-path/

Signed-off-by: Vagrant Cascadian <vagrant@reproducible-builds.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,6 +284,7 @@ CFLAGS		+=	$(GENFLAGS)
 CFLAGS		+=	$(platform-cflags-y)
 CFLAGS		+=	-fno-pie -no-pie
 CFLAGS		+=	$(firmware-cflags-y)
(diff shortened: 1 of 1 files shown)


alistair23 (approved):
Sounds good to me!

OpenSBI prefers patches be submitted to the mailing list for approval though, you can find more details about this here: https://github.com/riscv-software-src/opensbi/blob/master/docs/contributing.md

avpatel:
This has been taken care by new sbi_panic() and SBI_ASSERT() support in sbi_console.h

PR #232, opened 2021-11-27
docs/platform: spike: Improve examples

a4lg:
I mainly use Spike rather than QEMU for small tests and I find that setting up Spike (with OpenSBI + Linux + initramfs) is not easy enough.

This commit makes Spike examples usable as QEMU ones.
This is not a full tutorial (totally lacks initramfs, which is out of scope in the OpenSBI documentation) but better than nothing.


avpatel:
Can you please send your patch using "git send-email" to opensbi@lists.infradead.org ?

Regards,
Anup

a4lg:
Understood.

PR #235, opened 2022-01-10
Fix pointer of type 'void *' used in arithmetic

jlaitine:
Using "void *" in arithmetic causes errors with strict compiler settings:
"error: pointer of type 'void *' used in arithmetic [-Werror=pointer-arith]"

Just remove these by calculating on "char *" where 1-byte data size is assumed


avpatel:
We have mailing list based patch review and merging.

Can you please join OpenSBI mailing list and send patch using "git send-email" ?

To join OpenSBI mailing list, visit https://lists.infradead.org/mailman/listinfo/opensbi

Regards,
Anup

PR #237, opened 2022-01-11
Provide flexibility to specify OpenSBI FW_TEXT_START during compilation

weilianglim1785:
This changes is to provide flexibility for platform that does not use the default FW_TEXT_START.
If FW_TEXT_START is not specified, default value will be used.

Signed-off-by: Wei Liang Lim <weiliang.lim@linux.starfivetech.com>

### DIFF
--- a/platform/generic/config.mk
+++ b/platform/generic/config.mk
@@ -18,7 +18,9 @@ platform-runcmd = qemu-system-riscv$(PLATFORM_RISCV_XLEN) -M virt -m 256M \
   -nographic -bios $(build_dir)/platform/generic/firmware/fw_payload.elf
 
 # Blobs to build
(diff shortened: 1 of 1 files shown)


avpatel:
The OpenSBI firmwares are by default position independent so you can run it from any other text base without recompiling.

CodingVoid:
On current master branch, I can set FW_TEXT_START to tell OpenSBI where to link to. It is also an important feature in my opinion, because I avoid relocation whenever I can. It doesn't really take much perfomance, but debugging using GDB is very tedious if your program starts relocating itself to somewhere else.
So I guess the issue does not exist anymore?

avpatel:
No activity so assuming it is resolved.

PR #241, opened 2022-01-21
lib: utils/serial: Round UART8250 baud rate divisor to nearest integer

Limoto:
Previously, it was rounded down and that gives suboptimal results when
non-standard clock sources or baud rates are used.

Signed-off-by: Jakub Luzny <jakub.luzny@codasip.com>


avpatel:
Can you send your patch to opensbi@lists.infradead.org using GIT send-email ?

You will need to subscribe to this lists using following link:

Regards,
Anup

PR #242, opened 2022-01-21
lib: fix compilation when strings.h is included

pkarashchenko:
In a systems that provide `strings.h` and it is included together with `sbi_bitops.h` the compilation error appears.
The `ffs()` and `fls()` are provided by `strings.h`

Signed-off-by: Petro Karashchenko <petro.karashchenko@gmail.com>

### DIFF
--- a/include/sbi/sbi_bitops.h
+++ b/include/sbi/sbi_bitops.h
@@ -37,47 +37,12 @@
 	(((~0UL) - (1UL << (l)) + 1) & (~0UL >> (BITS_PER_LONG - 1 - (h))))
 
 /**
- * ffs - Find first bit set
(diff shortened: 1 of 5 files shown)


pkarashchenko:
As an alternative both `ffs()` and `fls()` can be removed from `sbi_bitops.h` as those are not used anywhere in the code

avpatel:
Can you send your patch to opensbi@lists.infradead.org using GIT send-email ?

You will need to subscribe to this lists using following link:

Thanks,
Anup

pkarashchenko:
Yes. Do you recommend to send the current patch or remove inline functions that have conflicts?

PR #246, opened 2022-02-14
fix detection of csr count

liangliang-ll:
The check of csr_swap() would always pass if '__wrval' is
the macro substitution of the variable 'val' in check_csr(),
which renders hfeatures->*_count incorrect.

Use 'tmp_val' instead of 'val' to keep the temporary value of
CSRs to avoid this problem.


avpatel:
Not applicable to latest OpenSBI.

PR #250, opened 2022-03-25
fixed a minor doc issue

unicornx:
"-append only allowed with -kernel option", which is required by qemu.

Signed-off-by: Wang Chen <wangchen20@iscas.ac.cn>


avpatel:

Regards,
Anup

PR #253, opened 2022-04-21
sbi: scale the stack according to pointer size

compnerd:
Scale the default stack size according to the pointer size.  This is
important to accommodate the space required for stack allocation of the
device tree structures which scale according to the pointer size.

In the case that the default stack size is insufficient, ensure that
there is an escape hatch for users to override the default stack size at
compile time.


avpatel:

Regards,
Anup

PR #259, opened 2022-06-11
Virtio serial driver

misterjdrg:
Using emergency_write configuration field in virtio mmio devices for early serial


avpatel:

Regards,
Anup

avpatel:
No activity on this hence closing.

PR #262, opened 2022-07-14
lib: sbi_hart: Shorten the code to set MPV bit

gagachang:
MPV bit is set when the value of next_virt boolean variable equals
true. Since the value of next_virt is either 0 or 1, we can set
MPV bit without if-else logic.

Signed-off-by: Che-Chia Chang <alvinga@andestech.com>


avpatel:
Overall you patch looks good.

Regards,
Anup

gagachang:
Since this patch has been reviewed by maling list and merged, I close it.

PR #276, opened 2022-11-19
Makefile: bugfix for handling platform paths

acaldaya:
If the path where this repo is located contains the platform name on it, the original Makefile replaced its occurrences from the path making it an invalid path. This commit prevents this behavior replacing only the last part of the path as originally intended.

Signed-off-by: Alejandro Cabrera Aldaya <aldaya@gmail.com>


avpatel:

Regards,
Anup

avpatel:
Already addressed.

PR #277, opened 2022-11-20
Makefile: replace `echo` with `printf` for compatibility

kadiwa4:
I don't know why but `echo -n` didn't work for me. macOS supports the `-n` option but it doesn't work in the makefile. What it does instead is it literally writes `-n` to the file and then also leaves a newline at the end.

I'm using GNU Make 4.4 (`gmake` from Homebrew).


avpatel:

Regards,
Anup

PR #295, opened 2023-03-18
fix: Fix the typos

howjmay:


wxjstz:
We have mailing list based patch review. You need to join mailing list: http://lists.infradead.org/mailman/listinfo/opensbi
Then send this patches use "git send-email"


Regards,
Xiang W

avpatel:

Regards,
Anup

Issue #291, opened 2023-03-13
How to calculate correct FW_JUMP_FDT_ADDR?


gsomlo:
For (at least) the generic platform, `FW_JUMP_FDT_ADDR` defaults to `0x0x82200000`:
https://github.com/riscv-software-src/opensbi/blob/master/platform/generic/objects.mk#L35

The recommended "algorithm" to determine if that's enough, shown in `docs/firmware/fw_jump.md`:
https://github.com/riscv-software-src/opensbi/blob/master/platform/generic/objects.mk#L35
seems broken/insufficient in my case.

I have built a 6.3.0-rc1+ kernel with Fedora's default config file at:
http://fedora.riscv.rocks:3000/rpms/kernel/src/branch/f37-riscv64/kernel-riscv64-fedora.config

Running the test on the resulting `vmlinux` file produces:

```
$ objdump -h linux/vmlinux | sort -k 5,5 | awk -n '/^ +[0-9]+ /{addr="0x"$3; size="0x"$5; printf "0x""%x\n",addr+size}' | tail -1
0x216ea20
```

which is clearly less than `0x2200000`.

I built opensbi with (DTS source: [nexys_video_fed.dts](https://github.com/riscv-software-src/opensbi/files/10958280/nexys_video_fed.dts.txt)):

```
make CROSS_COMPILE=riscv64-unknown-linux-gnu- PLATFORM=generic FW_FDT_PATH=~/DTS/nexys_video_fed.dtb
```

I'm loading:
 - `fw_jump.bin` to `0x80000000`
 - `Image-6.3.0-rc1+` to `0x80200000`
 - `initramfs-6.3.0-rc1+.img` to `0x83000000`

The result is a crash during early boot, which I have narrowed down to some of the tracing initialization callback addresses being clobbered in the kernel's memor range:

```
[    0.000000] trace event string verifier disabled
[    0.000000] rcu: Hierarchical RCU implementation.
[    0.000000] rcu:     RCU restricting CPUs from NR_CPUS=32 to nr_cpu_ids=4.
[    0.000000]  Rude variant of Tasks RCU enabled.
[    0.000000]  Tracing variant of Tasks RCU enabled.
[    0.000000] rcu: RCU calculated value of scheduler-enlistment delay is 10 jiffies.
[    0.000000] rcu: Adjusting geometry for rcu_fanout_leaf=16, nr_cpu_ids=4
[...]

PR #296, opened 2023-03-20
docs: Correct FW_JUMP_FDT_ADDR calculation example

gsomlo:
When using `PLATFORM=generic` defaults, the kernel is loaded at `FW_JUMP_ADDR = FW_TEXT_START + 0x200000`, and the FDT is loaded at `FW_JUMP_FDT_ADDR = FW_TEXT_START + 0x2200000`.

Therefore, the maximum kernel size before `FW_JUMP_FDT_ADDR` must be increased is the difference, or `0x2000000`.

Update the example calculation to reflect this, and avoid confusion.

Fixes issue #291


gsomlo:
[resubmitted to mailing list instead, here](https://patchwork.ozlabs.org/project/opensbi/patch/20230322170448.3600124-1-gsomlo@gmail.com/)

PR #304, opened 2023-05-06
Dereference symlinks when copying / installing files

filmil:
Adds the `-L` flag (follow symlinks) to the `cp` commands used to install `libsbi.a` and `include/sbi/*`.

This should make no difference in regular compilation. However, it does make a difference when compiling with bazel.  Namely, bazel's sandboxing will turn all the source files into symlinks. After installation with `cp` the destination files will be symlinks pointing to the sandbox symlinks - which are removed by default when compilation ends.

This means that, in bazel, the resulting include files will be unusable due to the dangling symlink issues. Adding `-L` when copying ensures that the files obtained by executing the `install` targets are always dereferenced to files, rather than symlinks, eliminating this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,10 +401,10 @@ merge_deps = $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     cat $(2) > $(1)
 copy_file =  $(CMD_PREFIX)mkdir -p `dirname $(1)`; \
 	     echo " COPY      $(subst $(build_dir)/,,$(1))"; \
-	     cp -f $(2) $(1)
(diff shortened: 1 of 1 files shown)


avpatel:

Regards,
Anup

filmil:
Hm, I'm not comfortable with poking a security hole on my email inbox for `git send-email`, since it only supports password login.

Do you have alternatives?

filmil:
Found a way to contribute the change, see http://lists.infradead.org/pipermail/opensbi/2023-May/004950.html

PR #311, opened 2023-06-22
DBCN: check A2 register in RV64.

glg-rv:
I am probably missing something, but in RV64 DBCN extension we ignore A2 register, i.e. the highest XLEN bits of the physical address.

While I do not expect to make the physical address a 128-bit register, nevertheless the spec (both section 3.2 and Chapter 12, I see no mention of RV64 not requiring to interpret A2 as a high bits of the physical address.

This PR fixes this by returning an error in case, similarly to RV32, the A2 is non-zero.


davidlt:
I would advice to send patches for a review to the mailing list. I don't think GH PRs are being used anymore.

PR #315, opened 2023-07-11
lib: sbi: Configure seed bits when MSECCFG is readable

sameo:
When MSECCFFG is not trapped, giving exclusive access to the SEED CSR to
S-Mode (SSEED=1, USEED=0) seems like a reasonable default. It gives the
Linux kernel the ability to add entropy to its randomness pool while
preventing user mode from accessing it.

Eventually, this check will be enhanced with a check for the Zkr
extensions through the riscv,isa-extensions dt-bindings.

### DIFF
--- a/include/sbi/riscv_encoding.h
+++ b/include/sbi/riscv_encoding.h
@@ -223,6 +223,9 @@
 #define ENVCFG_CBIE_INV			_UL(0x3)
 #define ENVCFG_FIOM			_UL(0x1)
 
+#define SECCFG_SSEED			(_ULL(1) << 9)
+#define SECCFG_USEED			(_ULL(1) << 8)
+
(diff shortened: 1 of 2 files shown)


avpatel:

Regards,
Anup

sameo:
> We have mailing list based patch review so it would be great if you can send these patchs to OpenSBI mailing list.
> 
> You need to join OpenSBI mailing list using following link http://lists.infradead.org/mailman/listinfo/opensbi

Thanks @avpatel. I am still waiting for my subscription request to get approved. Is this moderated?

Cheers,
Samuel.

avpatel:
Strange, there is no approval required

avpatel:
I have added your email manually to OpenSBI list. Let me know if you still face issue.

PR #322, opened 2023-08-24
Fix comment errors

zrmin:
First, When we call **_reset_regs**, it'll reset all registers except ra, a0, a1 and a2;
Second, both boot HART and non-boot HARTs will execute the **_start_warm** function. Therefore, **_reset_regs** called in **_start_warm** will reset all registers except ra, a0, a1 and a2 for boot HART and non-boot HARTs.


avpatel:

Regards,
Anup

PR #325, opened 2023-09-06
lib: sbi: Implement SET_FS_DIRTY() to make sure the mstatus FS dirty …

greentime:
…is set

We found the mstatus.FS status is not set correctly after the SET_F64_REG() and SET_F32_REG(). We should set mstatus.FS dirty after we emulate the FPU instructions.

Co-developed-by: Roy Lin <roy.lin@sifive.com>


davidlt:
Hey @greentime, could you send the patch to the mailing list? Basically all development is happening on it now.

PR #331, opened 2023-11-23
[RFC] Add Supervisor Software Events support

clementleger:
The SBI Supervisor Software Events (SSE) extensions provides a mechanism
to inject software events from an SBI implementation to supervisor
software such that it preempts all other supervisor level traps and
interrupts [1].

Various events are defined and can be send asynchronously to supervisor
software (RAS, PMU, DEBUG, Asynchronous page fault) from SBI as well
as platform specific events. Events can be either local (per-hart) or
global. Events can be nested on top of each other based on priority and
can interrupt the supervisor mode at any time.

This PR adds support for this extension. First commit modifies interrupts
handling to pass the sbi traps regs through interrupt handlers in order to inject
SSE events from this context. Second one add the SSE support itself.

A RFC for Linux which uses this extension is available at [2].

Link: https://lists.riscv.org/g/tech-prs/message/515 [1]
Link: https://lore.kernel.org/linux-riscv/20231026143122.279437-1-cleger@rivosinc.com/ [2]


lib/sbi/sbi_sse.c
```diff
@@ -0,0 +1,963 @@
...
+	struct sbi_sse_event *e;
+	const unsigned align = __riscv_xlen >> 3;
+	ulong smode = (csr_read(CSR_MSTATUS) & MSTATUS_MPP) >>
+			MSTATUS_MPP_SHIFT;
+
+	if (phys_lo & (align - 1))
+		return SBI_EINVALID_ADDR;
+
+	/*
+	 * On RV32, the M-mode can only access the first 4GB of
```
deepak0414, line 783: Shouldn't be an issue for RV64.
This gets compiled for RV64 too, right?

So `phys_hi` check should be under some `ifdef`, right?
Am I missing something here?

deepak0414 (commented):
I've a general comment on design of SSE which interferes with security. This is kind of a TOCTOU issue between kernel (less priv) and opensbi (more priv)

If a control flow integrity scheme (software or hardware assisted) is implemented to ensure control safety of kernel indirect branches, then SSE provides a mechanism using an attacker can exercise free branches. I believe software base kCFI support for riscv is already in mainline.

During time of event registration, kernel provides physical address of context. This context structure acts as 
-  Context to be used for event inject in S (and along with PC, SP and other registers)
-  Context to be used where interrupted context of S will be saved away so that it can be restored later once injected event has finished its course

This entire context buffer is mapped in kernel space and thus is vulnerable to memory corruption. If any CFI scheme is implemented, its expected that an attacker has "write what anywhere" bug. Thus this design allows an attacker to corrupt this context and use mechanisms to trigger events that will allow M mode to inject events in kernel and thereby allowing free branches (which CFI schemes disallow) in kernel.

Without understanding limitations of `opensbi`, I would like to propose a mitigation
- During registration, `opensbi` makes a copy of context structure
- Similarly interrupted context can also live in protected region in M mode memory

Let me know what you think.

PR #337, opened 2023-12-04
fix section type for firmware dynamic

mateoconlechuga:
This section is only intended to hold data, and should not be executable.

### DIFF
--- a/firmware/fw_dynamic.S
+++ b/firmware/fw_dynamic.S
@@ -129,7 +129,7 @@ fw_options:
 	REG_L	a0, (a0)
 	ret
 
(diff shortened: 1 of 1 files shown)


avpatel:

Regards,
Anup

mateoconlechuga:
Great, thanks - I should have read the contributing guide first :laughing:

mateoconlechuga:
Here is the patch: https://lists.infradead.org/pipermail/opensbi/2023-December/006041.html

mateoconlechuga:
Closing for now since the review is tracked through the mailing lists.

Issue #345, opened 2024-02-26
fw_start and fw_rw_offset not aligned


CodingVoid:
I pulled the recent master of OpenSBI and I now get the following error:
fw_start and fw_rw_offset not aligned

I checked and the error comes from here:
https://github.com/riscv-software-src/opensbi/blob/322b59847534cbbf484190b0957fb32bee36e2ec/lib/sbi/sbi_domain.c#L752

My values are as follows:
scratch->fw_start:   0x80020000
scratch->fw_rw_offset: 0x40000

I am not entirely sure why the alignment is checked between the offsets? And I would assume that my addresses for both segments are aligned just fine.

PR #346, opened 2024-02-26
lib: sbi_domain: Fix PMP condition for addresses

CodingVoid:
In order to put an address/size into an NAPOT PMP region, the address needs to be aligned to the size shifted two to the right.

Example:
fw_start: 0x80040000
fw_size:  0x00040000
PMP address: 0x80040000 | (0x00040000 >> 2)
fw_size is valid, but not in the current implementation

Fixes issue #345


avpatel:

Regards,
Anup

CodingVoid:
Sorry I didn't know that. I will send the patches via email.

PR #349, opened 2024-03-19
platform: generic: allwinner: sun20i-d1: Remove duplicated CSR definions

cmuellner:
All T-Head CSRs are already defined in `thead/c9xx_encoding.h`. Instead of redefining them with a slightly different name, let's reuse the values from there.


cmuellner:
Closing this PR: I will send this to the mailing list instead.

PR #352, opened 2024-04-05
create a variable for the ext data array size

rminnich:
On some platforms, opensbi fails as the loop in
fdt_parse_isa_one_hart is for 32 items,
and the array is only 15 at present.

### DIFF
--- a/lib/utils/fdt/fdt_helper.c
+++ b/lib/utils/fdt/fdt_helper.c
@@ -401,7 +401,7 @@ static int fdt_parse_isa_one_hart(const char *isa, unsigned long *extensions)
 				continue;			\
 			}
 
-		for (j = 0; j < SBI_HART_EXT_MAX; j++) {
+		for (j = 0; j < sbi_hart_ext_size; j++) {
 			set_multi_letter_ext(sbi_hart_ext[j].name,
 					     sbi_hart_ext[j].id);
 		}
--- a/include/sbi/sbi_hart.h
+++ b/include/sbi/sbi_hart.h
@@ -74,6 +74,7 @@ struct sbi_hart_ext_data {
 };
 
 extern const struct sbi_hart_ext_data sbi_hart_ext[];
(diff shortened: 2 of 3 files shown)


lib/utils/fdt/fdt_helper.c
```diff
@@ -401,7 +401,7 @@ static int fdt_parse_isa_one_hart(const char *isa, unsigned long *extensions)
 				continue;			\
 			}
 
-		for (j = 0; j < SBI_HART_EXT_MAX; j++) {
+		for (j = 0; j < sbi_hart_ext_size; j++) {
```
orangecms, line 404: Would this then clash with other iterations relying on `sbi_hart_ext` being that previous size?

E.g. just further down:
```C
	for (i = 0; i < BITS_TO_LONGS(SBI_HART_EXT_MAX); i++)
		extensions[i] |= hart_exts[i];
	return 0;
```

Can the creation of `sbi_hart_ext` be changed to have something like null entries? I don't really know C, would do this much different in Rust. :crab: Maybe just changing all the code would be better.

rminnich: yeah, in a modern language this would be quite different. I believe that the code you point to is unaffected, all this is doing is fixing an iterator.

rminnich: it lost my response. I believe that's ok, the issue here is the sbi_hart_ext array, not hart_exts or extensions.

orangecms (approved):
Following the little exchange, yea, looks good to me!

lib/utils/fdt/fdt_helper.c
rminnich: i.e. extensions is this
unsigned long extensions[BITS_TO_LONGS(SBI_HART_EXT_MAX)];
hart_exts is this

        hart_exts = sbi_scratch_offset_ptr(scratch, fdt_isa_bitmap_offset);

I think both are safe.

PR #356, opened 2024-05-15
sbi/sbi_types.h: Don't unconditionally define '__always_inline'

ElyesH:
This is to fix update opensbi to upstream build for coreboot. https://qa.coreboot.org/job/coreboot-gerrit/257449/testReport/junit/(root)/clang/EMULATION_QEMU_RISCV_RV64_/


avpatel:

Regards,
Anup

Issue #416, opened 2026-03-21
OpenSBI sbi_ecall_get_extensions_str() can overrun caller buffers when the extension list is longer than the destination


neosys007:
Title: OpenSBI sbi_ecall_get_extensions_str() can overrun caller buffers when the extension list is longer than the destination

Current upstream OpenSBI master still has a buffer-handling bug in lib/sbi/sbi_ecall.c. The helper sbi_ecall_get_extensions_str() builds a comma-separated list of SBI extension names into a caller-provided buffer, but it does not clamp the running offset before each append the way sbi_hart_get_extensions_str() already does. The code currently does this:

    sbi_snprintf(exts_str + offset, exts_str_size - offset, "%s,", t->name);
    offset = offset + sbi_strlen(t->name) + 1;

and afterwards it writes:

    if (offset)
        exts_str[offset - 1] = '';

The problem is that offset is increased based on the nominal name length, not on the actual number of bytes that fit in the destination buffer. If the caller supplies a buffer smaller than the total concatenated extension string, offset can move past exts_str_size. At that point exts_str_size - offset becomes negative and is passed to sbi_snprintf() as a large unsigned value, and the final ext[s]_str[offset - 1] write can also step past the caller buffer. In other words, this is not just string truncation; the offset arithmetic itself is unsafe once the list grows beyond the destination size.

For comparison, sbi_hart_get_extensions_str() in lib/sbi/sbi_hart.c already performs a guard of the form:

    if (offset + sbi_strlen(...) + 1 > nestr)
        break;

before calling sbi_snprintf(). sbi_ecall_get_extensions_str() does not do that check today.

Expected behavior: the helper should check the remaining capacity before each append and stop safely when the caller buffer is full.

Actual behavior: the helper can keep advancing offset beyond the buffer size and then hand an invalid length to sbi_snprintf(), which can lead to an out-of-bounds write in the caller buffer.

Yudis-bit:
Thanks @avpatel for reviewing and merging the fix in [f95648d3955d](https://github.com/riscv-software-src/opensbi/commit/f95648d3955d72f77e13315a990a6135303978a5). The caller-capacity guard and SBIUNIT redzone regression are now upstream together, addressing the buffer-handling issue reported here. I appreciate you taking this through the mailing-list review and into the tree.

PR #423, opened 2026-08-06
lib: utils/mailbox: clamp RPMI shmem queue name to RPMI_NAME_CHARS_MAX

Yudis-bit:
## Problem

`rpmi_shmem_transport_init()` copies a device-tree `reg-names` string into `qctx->name` (a `char[RPMI_NAME_CHARS_MAX=16]` buffer) using `sbi_memcpy(qctx->name, name, len)` without validating the length. If the device tree supplies a `reg-names` string longer than 15 characters, the copy overflows the fixed-size 16-byte buffer.

This is the same class of buffer-length issue as #416.

## Reproduction

A device-tree node with a `reg-names` entry longer than 15 characters (e.g., `"a2p-db-and-shmem"` at 17 bytes including null) will overflow the 16-byte `qctx->name` buffer.

## Fix

Add a length clamp before the `sbi_memcpy`:

```c
if (len >= RPMI_NAME_CHARS_MAX)
    len = RPMI_NAME_CHARS_MAX - 1;
sbi_memcpy(qctx->name, name, len);
qctx->name[len] = '\0';
```

The name is truncated if it exceeds 15 characters, and the buffer is always null-terminated.

## Test

The RPMI mailbox subsystem does not have a standalone unit-test suite, but the fix is validated by review against the existing pattern in the codebase (e.g., the `rpmi_shmem_transport_init()` function itself already validates the `reg-names` count against `RPMI_QUEUE_IDX_MAX_COUNT`).

## Limitations

- The `RPMI_NAME_CHARS_MAX` constant is defined in `include/sbi_utils/mailbox/rpmi_msgprot.h` as 16. The existing code already uses this constant for the buffer declaration.
- No new sbi_unit test is added because the RPMI subsystem currently has no unit-test infrastructure in the tree.
- Cross-compilation was not performed (requires RISC-V toolchain).

### DIFF
--- a/lib/utils/mailbox/fdt_mailbox_rpmi_shmem.c
+++ b/lib/utils/mailbox/fdt_mailbox_rpmi_shmem.c
@@ -663,7 +663,10 @@ static int rpmi_shmem_transport_init(struct rpmi_shmem_mbox_controller *mctl,
 		if (!name || (name && len < 0))
 			return len;
 
+		if (len >= RPMI_NAME_CHARS_MAX)
+			len = RPMI_NAME_CHARS_MAX - 1;
 		sbi_memcpy(qctx->name, name, len);
+		qctx->name[len] = '\0';
 
 		/* store the index as queue_id */
 		qctx->queue_id = qid;


Yudis-bit:
A correction to the description: `a2p-db-and-shmem` is 16 characters, and `fdt_stringlist_get()` returns a length excluding the terminating NUL. That example fills the 16-byte destination without terminating it; it does not itself demonstrate a write past the buffer. I should have distinguished those cases in the report.

The proposal bounds the stored name and explicitly terminates it. Would maintainers prefer truncating an overlong queue name, as proposed here, or rejecting it as invalid input?

The validation limitation in the description still applies: no cross-build or runtime validation is reported. I also note the mailing-list submission requirement in `docs/contributing.md`; this GitHub discussion does not replace that review.

